Issue #708, opened 2026-04-28
polish(tabs): three identical 'Orchestrator' tabs at top, selected one shows raw zsh terminal instead of Orchestrator empty state


hurttlocker:
## Repro

Fresh launch of v0.1.58 from `/Applications/o8.app`. After Next.js hydrates, the tab bar shows three tabs all labeled exactly **Orchestrator**, with the rightmost one selected. The selected tab renders a raw terminal session (\`marquisehurtt@iMac ~ %\` prompt), not the Orchestrator empty state.

## What's wrong visually

- Three tabs in the same tab strip with the same label is an instant tell that something is mislabeled — a designer would catch this in 0.1 seconds. Either the labels should be unique (\`Orchestrator\`, \`Orchestrator 2\`, \`Orchestrator 3\`) or duplicates should be deduped on restore.
- The selected tab shows a zsh prompt — that's a terminal, not an orchestrator. So either:
  - State restoration is putting the wrong \`kind\` on a restored \`TerminalTab\`, OR
  - The label "Orchestrator" is being applied to terminal tabs by default
- First tab has no close-X (correct, it's the pinned default), but tabs 2 + 3 have close-X — so they're user-spawned, just mislabeled.

## What's on screen (from screenshot)

\`\`\`
[• Orchestrator] [Orchestrator x] [Orchestrator x ←selected]
                                  marquisehurtt@iMac ~ %
\`\`\`

## Suspected cause

\`workspace-terminal\` tab restoration in \`src/components/desktop/workspace-terminal/\` is preserving \`kind: 'terminal'\` while the label-resolver uses the default \`Orchestrator\` string for any tab whose label is unset. Two paths break: (1) old user-created terminals saved without an explicit label, (2) the tile-layout migration in \`lib/tiles/operations.ts\` may be resetting labels on \`TILE_LAYOUT_VERSION\` bump but not the kind.

## Severity

P0 polish — first thing the user sees on a fresh launch is three identical tabs and a raw terminal. Cursor / Apple bar this would never ship.

## Acceptance

1. After reset (clear localStorage tile-layout) + relaunch, only the two pinned tabs (\`Orchestrator\` + \`Assistant\`) are present. ✓ should be the steady state.
2. User-spawned terminals get a default label of \`Terminal {N}\`, not \`Orchestrator\`.
3. Existing persisted state with mis-labeled tabs gets migrated to the correct label on load (one-time backfill).

## Found via

[...]

PR #716, opened 2026-04-28
fix(tabs): apply tab dedupe across all restore paths (closes #713)

### LINKED ISSUE #713: polish(tabs): #709 fix doesn't clear duplicate Orchestrator tabs on existing persisted state in v0.1.61
## Repro

User who had 3 "Orchestrator"-labeled tabs in v0.1.58. Force-installed v0.1.61 (which contains #709 fix from terminal-restore.ts). Relaunched. **Still see three tabs labeled "Orchestrator"**.

## What should have happened (per #709)

- `terminal-restore.ts` should drop persisted entries with `tabKind === 'orchestrator'` and let the controller inject a fresh single Orchestrator
- Surviving `kind: 'terminal'` tabs whose label is "Orchestrator" should get relabeled to `Terminal N`

## What's actually happening

Three tabs still labeled "Orchestrator" persist across the v0.1.61 launch. Either:
1. The fix is not running on the actual persisted file path
2. The fix runs but the rendered state is from a different store (e.g. webview localStorage independent of `~/.o8/terminal-states/`)
3. There's a second restore path that doesn't go through `terminal-restore.ts`
4. The selected/rightmost tab's kind is genuinely `'orchestrator'` and not falling into either the drop or relabel branch

## Acceptance

After installing v0.1.61, launching the app, no manual cache clearing — only one "Orchestrator" tab visible (pinned default), other previously-broken tabs become "Terminal 1", "Terminal 2", etc. Or are dropped entirely.

## Severity

P0 — the fix shipped, the user updated, the bug persists. The user has to manually wipe localStorage to recover.

## Related

- #708 (original duplicate-tab report)
- #709 (the fix that didn't fully take)

hurttlocker:
## Summary

The #709 fix only ran inside `computeRestoredTabs` on the client. It missed the disk file itself, the fallback read paths in the API route, and the migration effect couldn't dedupe — so users with broken persisted state (e.g. three "Orchestrator"-labeled tabs) kept seeing the bug after force-installing v0.1.61.

This PR puts the same drop-stale-orchestrator + relabel-terminal-with-Orch-label logic in the four places that matter so the file gets healed on first read, regardless of which code path opens it.

## Changes

- **`lib/terminal/tab-state.ts`** — new `sanitizePersistedTabs` + `sanitizePersistedState` helpers. Keep at most one `kind: 'orchestrator'` (only at index 0) and relabel surviving `kind: 'terminal'` tabs labeled "Orchestrator" to the next free `Terminal N`. Idempotent. Both `loadTabState` and `saveTabState` now apply it.

- **`app/api/panel/terminal-state/route.ts`** — new `readAndSanitizeStateFile` that reads, sanitizes, AND writes back to disk whenever the sanitizer mutated anything. Wired into every read path (primary file, canonical-scope redirect, legacy file, latest-non-empty fallback, latest-repo fallback). POST handler also sanitizes incoming payloads. This is the one-time migration step — any stale file heals on first GET.

- **`useWorkspaceTerminalController.ts`** — migration effect hardened: instead of "inject if missing", it keeps exactly one Orchestrator at index 0 and drops any duplicates. Idempotent.

The original `computeRestoredTabs` fix from #709 stays in place as defense-in-depth at the React state layer.

## Test plan

- [x] `npx tsc --noEmit` passes
- [x] Smoke-tested against user's reported broken state (3-orch from #708, 2-orch + terminal-with-Orch from #713) — sanitizes cleanly and is idempotent on re-application
- [x] Smoke-tested against clean states (Orch+Assistant, Terminal-only, Chat-only, Empty) — sanitizer is a no-op
- [x] `computeRestoredTabs` still emits correct labels on the user's 3-orch input (Terminal 1 + Terminal 2)
- [ ] Install build, relaunch with stale file on disk, verify no duplicate Orchestrator tabs visible
- [ ] Verify the on-disk file gets healed (orchestrator-kind entries dropped, terminal-with-Orch-label relabeled) on first launch

### DIFF
--- a/src/app/api/panel/terminal-state/route.ts
+++ b/src/app/api/panel/terminal-state/route.ts
@@ -4,7 +4,11 @@ import { NextResponse } from 'next/server';
 import { readFileSync, writeFileSync, mkdirSync, existsSync, readdirSync, statSync } from 'fs';
 import path from 'path';
 import { listRepos } from '@/lib/repos/registry';
-import { buildRepoStateScope } from '@/lib/terminal/tab-state';
+import {
+  buildRepoStateScope,
+  sanitizePersistedState,
+  type PersistedTabState,
+} from '@/lib/terminal/tab-state';
 
 const HOME = process.env.HOME ?? '/tmp';
 const STATE_DIR = path.join(HOME, '.o8');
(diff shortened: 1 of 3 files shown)


hurttlocker:
Superseded by the tab-identity campaign that landed on main — dedupe + restore-path correctness now live in the restore pipeline itself: 24ebe570 (packetId-first label matching + persisted mission-card dedupe), 435bf509 (archived-owned session reap at restore with positive-confirmation drop), plus the laneId retarget work in terminal-session-ops.ts. #713 is closed. Closing this rather than rebasing a 3-month-old branch.

Issue #740, opened 2026-04-28
feat: register codebase-memory-mcp in .mcp.json + setup config generator


hurttlocker:
## Summary
Register `codebase-memory-mcp` as a third MCP server in `.mcp.json` so any Claude Code session opening `~/cortex-ide` auto-loads its tools. Update `/api/setup/mcp-config` to emit the same registration for external Claude Desktop / Claude Code installs.

## Why now
Demo blocker — without registration, the orchestrator can't call `index_repository`, `trace_call_path`, etc.

## Acceptance criteria
- [ ] `.mcp.json` at repo root has a third entry `codebase-memory` pointing at the resolved binary path with stdio transport.
- [ ] `/api/setup/mcp-config` GET emits the same entry, resolving binary path per install mode (dev vs bundled).
- [ ] `/api/setup/claude-desktop` POST writes the entry to `~/Library/Application Support/Claude/claude_desktop_config.json` with merge-preserving logic + `.o8-backup-<ts>` sidefile.
- [ ] After a fresh Claude Code session in `~/cortex-ide`, tools appear under `mcp__codebase-memory__*` (verify with `mcp__o8__o8_view_eval` or `claude mcp list`).
- [ ] Settings → MCP tab shows the third server in the installed-server list.

## Implementation sketch
- Add `codebase-memory` block to `.mcp.json` with `command` resolved from `O8_CODEBASE_MEMORY_BIN`
- Update `src/app/api/setup/mcp-config/route.ts` to include third server in the JSON output
- Update `src/app/api/setup/claude-desktop/route.ts` POST handler to merge the third entry
- Update Settings → MCP tab UI in `src/components/desktop/settings/` to render third row
- Don't break the existing `o8` operator MCP entry — preserve order and merge logic

## Dependencies
- Vendor codebase-memory-mcp binary (#1 in epic)

## Out of scope
- Auto-indexing repos.
- Any orchestrator-side tool calls.
---
Part of #738

Issue #738, opened 2026-04-28
feat: context engine v2 — codebase-memory-mcp + recall card + dispatch hook


hurttlocker:
## Summary
Build o8's persistent code-aware context layer by bundling `codebase-memory-mcp` (DeusData/codebase-memory-mcp, MIT, single static binary, vendored SQLite, tree-sitter call graph for 66 languages) as a third MCP server, surfacing a Context Recall Card in the Orchestrator tab at packet dispatch, and injecting a `<context>` block into every dispatched packet body. Cortex v2 substrate (Directives + Session Ledger) stays — this epic adds a code-graph layer on top.

## Why now
YC demo on May 4 is the visible win — the Context Recall Card is the hero frame. The deeper play (auto-directive proposer, temporal decay, cross-repo learning) is the post-YC moat that no foundation lab will build because it's coding-fleet-specific governance, not generic chat memory.

## Framework alignment — Subramanya N's 6 jobs of context engineering

This epic implements the [6-job context engineering framework](https://subramanya.ai/2026/04/23/context-engineering-why-prompt-engineering-was-never-enough/) at fleet scale. Each child maps to one or more pillars:

| Pillar | Children | What we ship |
|---|---|---|
| **Selection** | #742 | Recall Card surfaces top relevant directive + last 3 outcomes per repo |
| **Structure** | #742 | Three Issues-style rows (Directive / Outcomes / Symbol graph) — predictable shape |
| **Compression** | #743 | `<context>` block capped at 4000 chars; symbol graph truncated first |
| **Memory** | #739 #740 #741 #745 #746 #748 | codebase-memory-mcp + temporal decay + auto-directive proposer + cross-repo |
| **Tool Design** | #743 | Hidden-from-transcript injection at runtime adapter boundary; per-tab toggle |
| **Isolation + Orchestration** | #747 | Per-runtime telemetry routing — Codex/Gemini/opencode get scored independently |

**Maturity-model positioning**: existing tools solve context engineering at Level 2–3 *for single agents*. o8 ships Level 3–4 *for fleets*, where isolation + orchestration become the dominant cost — that's the wedge no incumbent owns.

## Pre-YC track (must ship before May 4)
- [ ] #739 feat: vendor codebase-memory-mcp binary into Tauri sidecar
- [ ] #740 feat: register codebase-memory-mcp in .mcp.json + setup config generator
- [ ] #741 feat: auto-index repos at boot via codebase-memory-mcp
[...]

PR #754, opened 2026-04-29
feat(mcp): vendor codebase-memory-mcp into Tauri sidecar (closes #739)

### LINKED ISSUE #739: feat: vendor codebase-memory-mcp binary into Tauri sidecar
## Summary
Vendor the `codebase-memory-mcp` static binary (DeusData/codebase-memory-mcp, MIT) into `src-tauri/binaries/` for all 4 Tauri targets so the production app ships with code-graph indexing built in. Wire env vars so the bundled MCP server can resolve the binary path at runtime.

## Why now
Demo blocker — without the binary, no recall card, no dispatch injection.

## Acceptance criteria
- [ ] Binary present at `src-tauri/binaries/codebase-memory-mcp-{darwin-arm64,darwin-x64,linux-x64,windows-x64}` and committed via Git LFS or fetched at build time from a release artifact (pick one and document).
- [ ] If upstream lacks darwin-arm64 prebuilt: cross-compile from source, document the command in `docs/codebase-memory-build.md`.
- [ ] `src-tauri/src/lib.rs` resolves binary at startup and sets `O8_CODEBASE_MEMORY_BIN` env var for child processes.
- [ ] Binary executes in production build — verify with `mcp__o8__o8_view_screenshot` after install OR with a shell call to `${O8_CODEBASE_MEMORY_BIN} --version`.
- [ ] `npm run tauri:build:signed` succeeds with new binary included.
- [ ] Total app bundle size delta < 30 MB (codebase-memory-mcp is ~15 MB).

## Implementation sketch
- Download binaries from upstream releases or cross-compile via `cargo build --release --target ...`
- Add `src-tauri/binaries/.gitattributes` if using LFS
- Update `src-tauri/tauri.conf.json` `bundle.externalBin` array (or `bundle.resources` if not invoked directly)
- In `src-tauri/src/lib.rs`, after Node pre-flight: locate platform-correct binary, set `O8_CODEBASE_MEMORY_BIN`
- Update `scripts/tauri-export.mjs` if binary path needs to be referenced
- Test on signed dev build: `npm run ship` → install → confirm binary executes

## Dependencies
None (foundation issue).

## Out of scope
- Registering with `.mcp.json` (separate issue).
- Auto-indexing logic (separate issue).
- Any UI surface.
---
Part of #738

hurttlocker:
Closes #739. First child in the strict-sequential YC-prep Context Engine v2 chain (#738).

## Approach: Option B (real binary) via build-time fetch + single-arch ship

The issue offered two paths — write a thin Node MCP server (Option A) or vendor the real `DeusData/codebase-memory-mcp` static binary (Option B). The issue body strongly implies Option B (\"the codebase-memory-mcp static binary\", \"~15 MB\", `bundle.externalBin` references), so I went with the real binary.

Within Option B I picked **build-time fetch** over committing the binaries:

- Each architecture's binary is **~161 MB uncompressed**. Committing all four to LFS would push >640 MB through every clone and dominate LFS storage cost.
- A fetch script materializes only the matching-host binary into `out/server/` during `npm run tauri:prebuild`, so each installer carries exactly one architecture.
- SHA-256 checksums are pinned in the script and verified after download — tampered upstream artifacts get caught at build time.
- Fetch is non-fatal: offline runners log a warning and ship without the binary, and the Tauri sidecar treats a missing binary as \"feature unavailable\" rather than a startup error.

## What changed

- `scripts/fetch-codebase-memory.mjs` — new. Detects host platform, downloads + verifies + extracts, idempotent via a sentinel file.
- `scripts/tauri-export.mjs` — invokes the fetch script after the rest of the bundle is built.
- `src-tauri/src/lib.rs` — looks for `<resource_dir>/server/codebase-memory-mcp{.exe}`. If found, exports `O8_CODEBASE_MEMORY_BIN` to spawned children and to the parent process env (so any later child inherits it, mirroring the `O8_NODE_BIN` pattern).
- `docs/codebase-memory-build.md` — new. Documents the source, version pin, upgrade procedure, cross-compile fallback, size trade-off, and what downstream issues need.

No `tauri.conf.json` change required — the existing `bundle.resources` mapping (`out/server/` → `Contents/Resources/server/`) already ships everything in `out/server/`.

## Bundle size delta — flagging clearly

| Metric | Before | After (single arch) | Delta |
|---|---|---|---|
| Compressed installer (`.app.tar.gz`) | ~41 MB | ~69 MB | **+28 MB** |
| Installed `.app` on disk | ~148 MB | ~309 MB | **+161 MB** |

**This blows past the issue's `<30 MB` on-disk target and the user's ~220 MB total budget.** The issue body assumed ~15 MB; the real binary is ~161 MB because it ships tree-sitter parsers for 66 languages plus a UI bundle.

Two follow-up paths if disk delta turns into a blocker:

1. **Runtime download** — drop the build-time fetch, have the sidecar pull the binary into `~/.cortex-ide/codebase-memory-mcp/<version>/` on first launch. Zero installer delta, costs the user a one-time ~28 MB network pull. ~30 lines of Rust.
2. **Slimmed binary** — custom build with a reduced parser set. Out of scope for this issue but tracked in `docs/codebase-memory-build.md`.

Both are out of scope for #739 — flagged for orchestrator decision.

## Verification

- `npx tsc --noEmit` ✓ clean
- `cd src-tauri && cargo check --features dev-mcp-plugin` ✓ clean
- `node scripts/fetch-codebase-memory.mjs` ✓ first run downloads + extracts + writes binary, second run hits cache
- `out/server/codebase-memory-mcp --version` ✓ prints `codebase-memory-mcp 0.6.0` (verified on darwin-amd64)
- `cargo tauri build` not run locally (long-running + needs signing key); CI gate is the build trigger.

## Lifecycle / spawn behavior

The binary is **not auto-spawned** by `lib.rs` — that's #740's job. Today the sidecar just sets `O8_CODEBASE_MEMORY_BIN` so child processes can find the binary. When #740 wires up `.mcp.json` registration the Claude Code session will spawn it on demand and tear it down via the standard MCP stdio close pattern. The existing `register_child` / `kill_tracked_children` flow in `lib.rs` already handles SIGTERM/SIGINT cleanup for any future direct spawns.

## What #740 (next in chain) needs

- **Read `process.env.O8_CODEBASE_MEMORY_BIN`** at runtime. If set, register an MCP server entry in `.mcp.json` (or wherever the registration logic lives) with `command` pointing at that path and no args (the binary defaults to MCP-stdio mode when invoked bare).
- **If unset**, omit the entry entirely. Don't error.
- **Working directory** for the spawned binary should be the user's repo root — it indexes whatever directory it's invoked from. The orchestrator session already passes a `cwd` per session.

## Files touched

- `scripts/fetch-codebase-memory.mjs` (new)
- `scripts/tauri-export.mjs`
- `src-tauri/src/lib.rs`
- `src-tauri/Cargo.lock` (pre-existing version sync from main)
- `docs/codebase-memory-build.md` (new)

## Caveats / risks

- **Size delta** is the headline. See above — orchestrator decides whether to ship as-is or pivot to runtime download.
- **CI build hosts must have network access** to `github.com/DeusData/codebase-memory-mcp/releases`. If a network-restricted runner is added later, switch to runtime download.
- **Upstream pin is v0.6.0**. Bump procedure in `docs/codebase-memory-build.md` — mainly bump `CMM_VERSION` and refresh the `CHECKSUMS` map.
- **Windows** path is wired (`.exe` extension, zip extraction via `unzip` / `Expand-Archive`) but **not smoke-tested on a Windows host** — only macOS x64 verified end-to-end here.

### DIFF
--- a/docs/codebase-memory-build.md
+++ b/docs/codebase-memory-build.md
@@ -1,0 +1,68 @@
+# Vendoring `codebase-memory-mcp` (issue #739)
+
+The Context Engine v2 (epic #738) ships the static `codebase-memory-mcp` binary inside the o8 Tauri bundle so the production app can index repos and answer recall queries without a separate install step. This doc explains how the binary gets vendored, how to upgrade the pin, and the size trade-off the orchestrator should be aware of.
+
+## Source
+
+- Upstream: [DeusData/codebase-memory-mcp](https://github.com/DeusData/codebase-memory-mcp) (MIT)
+- Pinned version: `v0.6.0` (see `scripts/fetch-codebase-memory.mjs` `CMM_VERSION`)
+- Distribution: prebuilt static binaries on the GitHub release page, four targets — `darwin-amd64`, `darwin-arm64`, `linux-amd64`, `linux-arm64`, plus a `windows-amd64` zip.
+
+## How it ends up in the bundle
+
+1. `npm run tauri:prebuild` runs `scripts/tauri-export.mjs`.
+2. After the Next.js + MCP server bundles compile, the script invokes `node scripts/fetch-codebase-memory.mjs`.
+3. The fetch script detects the host platform (`process.platform` + `process.arch`), downloads the matching tarball/zip from `github.com/DeusData/codebase-memory-mcp/releases/v0.6.0`, verifies the SHA-256 against the pinned checksum, extracts the binary, and writes it to `out/server/codebase-memory-mcp` (or `.exe` on Windows).
+4. A sentinel file `out/server/.codebase-memory-mcp.version` marks the version + asset combo so reruns short-circuit.
+5. Tauri's `bundle.resources` config (`tauri.conf.json` already maps `out/server/` → `Contents/Resources/server/`) ships the binary alongside the Node servers — no `tauri.conf.json` change required.
+6. At launch, `src-tauri/src/lib.rs` looks for `<resource_dir>/server/codebase-memory-mcp{.exe}`. If found, it sets `O8_CODEBASE_MEMORY_BIN` on the spawned Next.js server's environment (and on the parent's env so any later child inherits it).
+
+## Why a fetch script instead of committing the binary
+
+Each architecture's binary is ~161 MB uncompressed. Committing all four (or even one) would either bloat every clone or push us into Git LFS storage costs. The fetch script keeps the repo clean and only ever materializes the binary for the current build host.
+
+The fetch is **non-fatal** — if the runner is offline, the prebuild logs a warning and continues. The Tauri sidecar treats a missing binary as "feature unavailable" rather than a startup error, so a bundle without the binary still boots; it just can't serve the codebase-memory MCP tools.
+
+## Bundle size impact
+
+| Metric | Before | After (single arch) | Delta |
+|---|---|---|---|
+| Compressed installer (`.app.tar.gz`) | ~41 MB | ~69 MB | +28 MB |
+| Installed `.app` on disk | ~148 MB | ~309 MB | +161 MB |
+
+This **exceeds** the issue's stated `<30 MB` delta target on disk. The compressed download delta is in the same neighborhood as the issue's estimate (the binary compresses 5.7×). Two follow-up paths if the disk delta becomes a problem:
+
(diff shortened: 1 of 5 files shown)


hurttlocker:
Holding this PR — user opted for runtime-download (Option 2) instead of build-time vendoring due to +161 MB on-disk delta. Refactor tracked in #755 on a new branch. This PR will be closed when #755 lands.

hurttlocker:
Superseded by #756 (runtime-download model preserves the 148 MB on-disk baseline). Closing per #755 spec.

Issue #1067, opened 2026-05-15
#1066.1 — Persistent Claude Code stream-json session module


hurttlocker:
Parent: #1066 (epic #1051).

Create `src/lib/claude-code/interactive-session.ts` — a per-chat-tab persistent `claude` session, modeled on `src/lib/lane/codex-orchestrator-session.ts`.

- Spawn `claude --input-format stream-json --output-format stream-json --verbose` — **NO `-p`** (that flag is the programmatic-billing line; dropping it keeps the session on the user's Max subscription — verified live: `apiKeySource:none`, `rateLimitType:five_hour`).
- Keep stdin open; feed each user message as one NDJSON line `{"type":"user","message":{"role":"user","content":"..."}}`. The process stays alive and accumulates context across turns.
- `Map<tabId, ClaudeCodeInteractiveSession>` registry: `{ proc, sessionId, status: ready|busy|dead, stdinWritable, createdAt }`.
- `ensureSession(tabId, cwd, model)` spawns if absent/dead; `sendMessage(session, text, onEvent)` writes the NDJSON line and routes stdout through the existing `createClaudeCodeStreamJsonParser` (`src/lib/claude-code/stream-json-parser.ts`).
- Lifecycle: kill on tab close, idle timeout, abort signal — copy the abort/timeout plumbing from `codex-orchestrator-session.ts`.
- Capture `session_id` from the `init`/`result` events for later `--resume`.
- Empirically test whether `--include-partial-messages` works without `-p` (help says it needs `-p`, but that help text is unreliable); fall back to whole-message events if rejected.
- Do NOT use `src/lib/terminal/` PTY infra — this is clean stdio.

`npx tsc --noEmit` must pass. Foundation — no dependencies.

hurttlocker:
Merged into main — persistent interactive Claude Code stream-json session module (src/lib/claude-code/interactive-session.ts).

Issue #1066, opened 2026-05-15
Claude Code chat tab: drive interactive `claude` REPL (PTY) instead of `claude -p` — bill subscription, not programmatic credits


hurttlocker:
**ARCHITECTURAL — needs an architect/brainstormer pass before dispatch. Do NOT auto-dispatch via the /loop.**

Parent: epic #1051. Resolves the billing premise raised in #1060.

## Why

The Claude Code chat tab restored in #1059 spawns `claude -p` (headless / print mode) — see `src/app/api/claude-code/send/streaming-spawn.ts:31`. Per Anthropic's June 15 2026 billing split, the `-p` flag itself routes usage to the separate **Agent SDK programmatic credit pool** ($20–200/mo), even with subscription OAuth auth. Interactive Claude Code (REPL/terminal/IDE) stays on the Pro/Max subscription.

So the chat tab today silently burns programmatic credits. The intent of #1051 is "the real Claude Code, in a tab" — it should bill the user's subscription like the Claude Code terminal does.

**This is what Superconductor and Conductor do:** both spawn the interactive `claude` REPL as a subprocess (no `-p`), using the user's subscription OAuth. That is the only path that bills Pro/Max.

## The work — component parts

> NOTE (filed by orchestrator): this decomposition is a first pass. @marquisehurtt reviewed Superconductor's component structure separately — reconcile this list against that review before splitting into child issues.

1. **PTY session manager** — spawn interactive `claude` (no `-p`) under a PTY, one persistent session per chat tab. Lifecycle: start / send / resume (`--continue` / `--resume`) / interrupt / exit. Reuse o8's existing PTY infra in `src/lib/terminal/`.
2. **REPL I/O protocol layer** — THE HARD PART. Interactive `claude` renders an ANSI TUI, not clean `stream-json`. Determine the cleanest programmatic channel: does interactive mode support `--input-format stream-json` / `--output-format stream-json` without `-p`? If not, define how output is parsed. This question should be answered first, in the architecture pass.
3. **Event bridge** — map REPL output into the existing SSE event vocabulary (`delta`, `thinking`, `tool_call`, `tool_result`, `usage`, `plan_step`, `permission_request`) so the parser (#1061) and renderers (#1062) stay valid and need no rework.
4. **Adapter + route rewire** — `/api/claude-code/send` and `src/lib/runtimes/claude-code.ts` move from spawn-per-message to a persistent REPL session keyed per tab.
[...]

PR #1072, opened 2026-05-15
#1066.1 — Persistent Claude Code stream-json session module

hurttlocker:
## Summary
- Add `src/lib/claude-code/interactive-session.ts` as the per-chat-tab persistent Claude Code stdio session registry.
- Spawn `claude --input-format stream-json --output-format stream-json --verbose --include-partial-messages` without `-p` or `--print`, keep stdin open, and send user turns as NDJSON.
- Route stdout through `createClaudeCodeStreamJsonParser`, capture `session_id` from `init`/`system`/`result` events, and expose tab-close/idle/abort cleanup hooks.

## Partial-message flag check
- Empirical local check accepted `--include-partial-messages` without `-p` on the resolved Claude binary. The probe exited `0`, emitted stream-json system hook events, and produced no stderr, so this module includes the flag.

## Verification
- `npx tsc --noEmit`
- `npm run typecheck`
- `npx eslint src/lib/claude-code/interactive-session.ts`
- `git diff --check`
- Static self-review with `git diff main...HEAD`

Refs #1067

### DIFF
--- a/src/lib/claude-code/interactive-session.ts
+++ b/src/lib/claude-code/interactive-session.ts
@@ -1,0 +1,455 @@
+import 'server-only';
+
+import { spawn } from 'node:child_process';
+import type { ChildProcessWithoutNullStreams } from 'node:child_process';
+import { homedir } from 'node:os';
+import { join, resolve } from 'node:path';
+import {
+  createClaudeCodeStreamJsonParser,
+} from './stream-json-parser';
+import type {
+  ClaudeCodeStreamJsonParser,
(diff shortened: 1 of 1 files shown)


hurttlocker:
Superseded by mainline: the persistent Claude Code stream-json session shipped via `src/lib/lane/orchestrator-session.ts` (interactive REPL spawn, never `claude -p`) and the warm REPL pool (`src/lib/claude-code/warm-repl-pool.ts`, d6d4e92a). #1066 is closed. Closing this branch rather than rebasing a 2-month-old parallel implementation.

Issue #1179, opened 2026-06-07
Review gate: independent second-pass reviewer with fail-closed AGREE-gate (from #1154)


hurttlocker:
Split from #1154 (review-gate hardening). #1154 ships the low-risk slice (structured emit-format heuristics, a deterministic warn-tier partition check, behavior-inspection prompt, and the high-risk classifier extension). This tracks the **structurally hardest, highest-impact** piece, deferred so it can ship deliberately with the manual governance benchmark watching the false-positive rate.

## The gap it closes
Today's "adversarial second pass" is a **same-turn, same-model** instruction ("do a SECOND skeptical pass in THIS SAME turn", `review-risk.ts`). A model that rationalizes a bug on pass 1 rationalizes it again on pass 2 of the same turn — which is exactly how the scope-partition diff slipped. The fix is a genuinely **independent** second reviewer that must AGREE before approve.

## Design (from the #1154 architecture map)
- Gate the second pass on `classifyReviewRisk().tier === 'high'` AND first-pass-approve (cost ceiling — only the small slice touching live state / data writes).
- Run it as a **separate `backend.sendTurn`** with a **blind** prompt (do NOT show pass-1 reasoning — only the diff + structured findings — or it anchors and rubber-stamps). "A prior reviewer approved this; find the partition leak / inert guard / missing sub-requirement they missed."
- **Enforce agreement (Option A, fail-closed):** add `requiresSecondPass`/`secondPassAgreed` to the `orchestrator_review` approval args (`submitPacketReview`, `review.ts`); tighten `hasDurableApprovedReview` (`durable-review-approval.ts`) so a `requiresSecondPass:true` row only counts as approved once `secondPassAgreed === true`. Keeps the single-chokepoint invariant; first-pass approve becomes provisional → operator card until agreement lands.

## Caveats to respect
- **Backend asymmetry (#1045):** the Codex orchestrator backend can't call `submit_review`/`lane_command`, so it can't stamp agreement — on that path the second pass is advisory and the operator card is the backstop. Don't design the AGREE-gate assuming Codex can stamp it. Fail-closed on both paths.
- **False-positive risk:** watch `fpRate` in the governance benchmark; intentional global writes (e.g. the directive store) are legitimate.
- Add a 7th case to `scripts/smoke-merge-requires-review.ts`: high-risk lane + first-pass approve + no second-pass agreement must NOT merge.

## Acceptance
[...]

Issue #1154, opened 2026-06-02
Harden the orchestrator review gate — benchmark showed it confidently approves subtly-broken diffs


hurttlocker:
The governance review→refix benchmark (2026-06-02) is the moat test — and it found a measured hole. The AI review tier caught 2/3 injected bugs but MISSED one by looking directly at the defective behavior and rationalizing it as "VERIFIED correct" (a scope-partition/duplication bug). All three bugs passed `tsc` + `lint`, so an ad-hoc merge-on-green path shipped 3/3 broken; o8 caught 2. The review gate is our differentiator — this is a real, reproducible weakness in it.

## Proposed
- Add explicit review heuristics for the failure classes that slipped past: scope/partition duplication (output written to the wrong/too-many places), inert/no-op guards (verify a guard's predicate against *actual* codebase behavior, not its name), and missing-sub-requirement coverage (does the diff satisfy ALL enumerated requirements?).
- For high-risk diffs (live state machines, data writes), run a second independent reviewer pass (adversarial-verify) and require agreement before approve.
- Prefer inspecting actual output/behavior over reading the diff where feasible.

## Acceptance
Re-run the governance benchmark: catch rate improves above 2/3 without raising false positives above 0/2. The human approval gate remains the backstop; this hardens the AI tier beneath it.

hurttlocker:
Done — orchestrator review gate hardened: `1699f1d0` (heuristics, #1180) + `cb43dc68` (second-pass agreement for high-risk diffs, #1186), wired via `blind-second-pass-review.ts` in `auto-review.ts`. Both proposed mechanisms shipped. Closing.

PR #1180, opened 2026-06-07
fix: harden orchestrator review-gate heuristics (#1154)

hurttlocker:
Addresses #1154 (the independent-second-pass AGREE-gate is split to #1179).

The governance benchmark found the AI review tier confidently approved a subtly-broken diff (scope-partition/duplication) by rationalizing the defect as "VERIFIED correct." The heuristic *traces* already existed in the review prompt but were prose a model can reason past. This PR makes the slip-classes harder to rationalize and adds a deterministic signal the prompt can't talk its way around.

## Changes (low-risk slice)
**1a — structured, mandatory emit-format** (`auto-review.ts` `buildReviewPrompt`): the verification traces are now a required machine-checkable block — `SCOPE: <file:line> partition=<key|NONE>` (+ single-destination / no-duplication check), `GUARD: <file:line> fires-from=<…|INERT>`, and a per-sub-requirement `COVERAGE` checklist. Hard rule: **may not approve** with any INERT guard, `partition=NONE` write, or unchecked box; submit_review instruction tightened to "approved only if all traces clear." EXECUTION-PATH trace retained.

**3a/3b — behavior inspection**: "where the change has observable output, PREFER to run it in the worktree and inspect the ACTUAL output over reasoning about the diff" + the worktree path is now surfaced in the prompt (threaded through `getDiffSummary.cwd`). You can't rationalize where a file actually landed.

**1b — deterministic warn-tier partition check** (`merge-gate.ts` `checkScopePartitionHeuristics`): flags added `writeFile/mkdir/INSERT|UPDATE/.set|.put` lines lacking a partition token. **`severity: 'warn'` — never blocks** (FP-safe); it surfaces the finding into the review prompt via the existing `formatMergeGateForReview` path. Verified it cannot escalate: `checkSelfReviewIntegrity` only counts `block`-severity violations.

**Risk classifier extension** (`review-risk.ts`): adds live-state-machine paths (`lane/{commands,reconcile,lifecycle,worktree-side-merge}.ts`, `orchestrator/*{state,store,control-plane}*`) + a `setLaneStatus` line rule to the high-risk tier. Patterns are anchored/tight to avoid over-matching o8's own dogfood traffic. Sets up #1179.

## Verification
- `npx tsc --noEmit` → exit 0.
- Merge-gate preview: `wouldMerge: true`, all checks pass.
- Surgical: 3 files (+100/−13).
- **FP-safety**: the new partition check is `warn` only and cannot escalate to a block.

**⚠️ The governance benchmark is manual** (no automated runner — `tests/bench/`, `npm run bench:score`). Re-run the 5-diff governance set (must include the scope-partition diff that slipped) before/after merge to confirm `catch_rate` improves without `fpRate` rising past the ±0.05 band. The deferred independent second-pass (#1179) is the piece that structurally closes the single-reviewer-rationalizes hole.

Built by a Codex worker in an isolated worktree; orchestrator-reviewed. Not merged — PR-only dogfood mode.

### DIFF
--- a/src/lib/lane/auto-review.ts
+++ b/src/lib/lane/auto-review.ts
@@ -238,6 +238,7 @@ interface ReviewDiffSummary {
   summary: string;
   changedFiles: string[];
   addedLines: string[];
+  cwd: string;
 }
 
 function extractAddedLines(diff: string): string[] {
@@ -273,18 +274,20 @@ function getDiffSummary(lane: Lane, depth: ReviewDepth): ReviewDiffSummary {
     const addedLines = extractAddedLines(diff);
 
     if (!stat && !diff) {
-      return { summary: 'No changes detected in the worktree.', changedFiles, addedLines };
+      return { summary: 'No changes detected in the worktree.', changedFiles, addedLines, cwd };
     }
     return {
       summary: `## Diff summary\n\n\`\`\`\n${stat}\n\`\`\`\n\n## Changes\n\n\`\`\`diff\n${diff}\n\`\`\``,
       changedFiles,
       addedLines,
+      cwd,
     };
   } catch {
     return {
       summary: 'Unable to generate diff — the worktree may not have commits yet.',
       changedFiles: [],
       addedLines: [],
+      cwd,
     };
   }
 }
@@ -434,6 +437,7 @@ function buildReviewPrompt(
   mechanicalChecksSummary?: string,
   mergeGateResult?: MergeGateResult,
   reviewScreenshot?: LaneReviewScreenshotReference | null,
+  reviewWorktreePath?: string,
 ): string {
   const depthGuidance = depth === 'deep-dive'
     ? 'This lane has low-confidence or missing self-review context. Perform a deep-dive review and challenge assumptions, edge cases, and missing validation.'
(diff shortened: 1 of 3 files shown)


hurttlocker:
## Moat validation (ran the hardened gate against the bug class that slipped)

Built a throwaway-git-repo harness and ran the **actual** `runMergeGate` / `classifyReviewRisk` / `hasScopePartitionToken` from this branch against a reconstructed scope-partition/duplication bug (an unpartitioned write to a ledger/store path — the #1154 miss class). **10/10 checks pass:**

- ✅ The unpartitioned write now produces a **deterministic merge-gate `warn`** ("File write without partition token") — the rationalization-proof signal that's force-fed into the review prompt via `formatMergeGateForReview`, independent of model judgment.
- ✅ It's **`warn`, never `block`** — and a properly-spaced partitioned write (`join(dataDir, scope, …)`) does NOT trip it (no false positive on clean diffs).
- ✅ `classifyReviewRisk` tiers the bug **HIGH** → fires the adversarial protocol + the new structured "may NOT approve if any write has partition=NONE" emit-format.
- ✅ The new state-machine path rule tiers `lane/commands.ts` (`setLaneStatus`) HIGH.

### What this proves vs. doesn't
**Proven:** the *structural* hardening works — the slip-class now generates a deterministic, non-rationalizable finding + high-tier escalation that always reaches the reviewer (and the operator card). This is the real fix vs. the prose-only traces that got rationalized.
**Not proven here (needs the manual review→refix benchmark):** whether the LLM reviewer now *rejects* the diff. The deterministic warn means the finding surfaces regardless, but the full catch-rate number requires a live run. The independent second-pass (#1179) is what closes the model-judgment half.

### ⚠️ Found a real FP edge — recommend a follow-up
The partition-token regex `\b(?:repo|tenant|user|project|lane|packet|scope|slug|id)\b` uses **word boundaries**, so it MISSES partition tokens buried in camelCase: `writeFile(perRepoLedgerPath(repoId), …)` is genuinely partitioned but gets flagged (`repoId` / `perRepoPath` have no `\b` around "repo"/"id"). Warn-only so it can't block, but it'll add noise on camelCase code (most of o8). Suggest a camelCase-aware token match (or splitting identifiers on case boundaries before the test) as a refinement — and watch `fpRate` in the governance benchmark.

PR #1196, opened 2026-06-07
fix(review-gate): match camelCase partition tokens to cut warn FP noise (#1184)

### LINKED ISSUE #1184: review-gate: partition-token check misses camelCase tokens (warn FP noise)
Follow-up from validating #1180 (review-gate hardening). The deterministic partition check works on the target bug class, but has a false-positive edge on camelCase.

## The edge
`hasScopePartitionToken` (`src/lib/lane/review-risk.ts`) uses `\b(?:repo|tenant|user|project|lane|packet|scope|slug|id)\b`. Word boundaries mean partition tokens **buried in camelCase identifiers are missed**, so a genuinely-partitioned write gets flagged as unpartitioned:
- `writeFile(perRepoLedgerPath(repoId), data)` → flagged (FP) — `repoId` / `perRepoPath` have no `\b` around "repo"/"id".
- `writeFile(join(dataDir, scope, 'x.json'), data)` → correctly clean (token is `\b`-visible).

Warn-only, so it never blocks a merge — but it adds noise on camelCase code (most of o8), which erodes the signal of the new `checkScopePartitionHeuristics` warn (`merge-gate.ts`).

## Proposed
Make the check camelCase-aware without reintroducing false negatives (the `\b` exists to avoid matching "id" inside "considered"). E.g. case-split the line first, then test:
```ts
const decamel = line.replace(/([a-z0-9])([A-Z])/g, '$1 $2');
return PATTERN.test(line) || PATTERN.test(decamel);
```
This catches `perRepoPath`→"per Repo Path", `repoId`→"repo Id" while still not matching "id" inside lowercase words. Note it won't catch stemmed forms (`scopedPath`→"scoped" ≠ "scope") — accept that or add explicit stems. Watch `fpRate` in the governance benchmark after.

## Validation reference
A throwaway-git harness ran the real `runMergeGate`/`classifyReviewRisk`/`hasScopePartitionToken` — 10/10 on the core behavior; this FP edge is the one documented gap (see PR #1180 comment).

hurttlocker:
Closes #1184.

`hasScopePartitionToken` matched only whole-word tokens (`\b(repo|tenant|user|project|lane|packet|scope|slug|id)\b`), which **misses camelCase identifiers** — `\buser\b` does not match inside `userId` (no word boundary between `r` and `I`). The scope-partition write heuristic at `merge-gate.ts:347` treats a present partition token as 'properly scoped' and **suppresses** the advisory warn, so missing camelCase tokens produced false-positive warnings on correctly-scoped writes (`repoId`, `tenantId`, `laneId`, `projectSlug`, …).

**Fix:** also run the pattern against a camelCase-split copy of the line (`userId` → `user Id`), so those identifiers register and the FP warn is suppressed.

- Single file: `src/lib/lane/review-risk.ts` (+5/-1)
- `tsc --noEmit` → 0; no test runner in repo. Consumer: `merge-gate.ts` (suppression direction confirmed).

### DIFF
--- a/src/lib/lane/review-risk.ts
+++ b/src/lib/lane/review-risk.ts
@@ -32,9 +32,13 @@ const ADVISORY_LINE_RULES: Array<{ pattern: RegExp; reason: string }> = [
 ];
 
 const SCOPE_PARTITION_TOKEN_PATTERN = /\b(?:repo|tenant|user|project|lane|packet|scope|slug|id)\b/i;
+const CAMEL_CASE_BOUNDARY_PATTERN = /([a-z0-9])([A-Z])/g;
 
 export function hasScopePartitionToken(line: string): boolean {
-  return SCOPE_PARTITION_TOKEN_PATTERN.test(line);
+  return (
+    SCOPE_PARTITION_TOKEN_PATTERN.test(line)
+    || SCOPE_PARTITION_TOKEN_PATTERN.test(line.replace(CAMEL_CASE_BOUNDARY_PATTERN, '$1 $2'))
+  );
 }
 
 function addUnique(reasons: string[], seen: Set<string>, reason: string): void {


hurttlocker:
Superseded by c205fad9 (`fix: detect camelcase partition tokens (#1195)`), which landed the same camelCase handling on main — `hasScopePartitionToken` in `src/lib/lane/review-risk.ts:37` now retests after a camelCase-boundary split. #1184 is closed. This branch's version never merged; closing.

PR #1603, opened 2026-07-21
feat: add composer undo-send buffer

### LINKED ISSUE #1240: Revamp default-side composer send UX to match canvas (undo-send buffer + queue) — once canvas is stable
## Why

The canvas now has a really smooth "humans make mistakes" layer on **every** composer (bottom stage composer, the dock, and floating chat cards). It feels great in use — we want the **default IDE side to eventually match it**. Saving this now so we don't lose the reference; the actual revamp happens **once canvas is stable** (don't start until then).

Two behaviors, identical across every canvas composer:

1. **Undo-send grace buffer.** After a message goes out, a quiet **"↩ Undo send"** pill appears for ~4.5s with a depleting underline (so you can see the window closing). Tap it — or hit Stop on the bottom composer — and it halts the run, snaps the text **+ images** back into the composer, and erases the transcript trace, as if you never sent it. Catch it after the window and Stop just stops (the send stands).
2. **Queue when busy.** Send while the orchestrator is mid-turn and the message queues (small cancelable pills) instead of being blocked/dropped; the head fires automatically when the current turn settles (edge-detected so they serialize one at a time).

## Reference implementation (the new gold standard)

- `src/app/preview/canvas-glass/use-send-buffer.tsx` — the shared `useSendBuffer` hook + `UndoSendPill` + `QueuedSends` components. Conversation-agnostic; the host wires in `dispatch` / `interrupt` / `restore` / `truncate`.
- Wired into all three surfaces:
  - Bottom composer — `page.tsx` (`dispatchMain` → `mainSend` / `mainStopOrUndo`; Stop button does undo-within-grace).
  - Dock — `dock.tsx` (queue + undo pill above the reply composer; composer now lets you type/send while busy).
  - Chat cards — `chat-card.tsx` (per-card buffer).
- Added `interrupt()` to `useThreadOrchestrator` in `use-canvas-orchestrator.ts` — per-thread chat cards literally couldn't stop a running turn before this.
- Undo truncation mechanism: capture `entryIdRef.current` as the boundary at dispatch time; on undo, filter the lane's entries to `id < fromEntryId` (removes user bubble + thinking + any partial response cleanly).

## Default side today (what we'd compare against / replace)

- Queue = `pendingSteers` in `src/components/desktop/thoughts/ThoughtsChatPanel.tsx` + `src/components/desktop/thoughts/chat-panel/useSteerAutoFire.ts` — steer-specific, **text-only**, rendered as editable rows in `ComposerArea.tsx`.
- **No undo-send grace buffer exists on the default side at all.**

## The ask (later)

- [ ] Compare behavior + look: canvas pill/queue vs default steer rows.
- [ ] Port the **undo-send grace buffer** to the default composer (it has none today).
- [ ] Unify/replace the default queue with the canvas's cleaner model — ideally share `useSendBuffer`. Note the seam: the canvas owns convo entries in page state (truncation is trivial there), while the default side has its own transcript store, so a literal share likely needs a small adapter for the `dispatch`/`truncate` callbacks.
- [ ] Keep **image** support (canvas bottom composer queues/undoes images; the default `pendingSteers` queue is text-only).

## Not now

Tracking-only. Do this **after** canvas stabilizes. Canvas implementation is currently uncommitted (landing with the broader canvas work).

hurttlocker:
Closes #1240.

Adds canvas-parity send handling to the default desktop composer: a 4.5-second undo window, text and image restoration, durable transcript truncation, late-token suppression, and a serialized cancelable queue. Canvas and desktop now share the headless composer send-buffer state machine.

Validation:
- npx tsc --noEmit
- focused Vitest coverage: 52/52 passed
- npm run rule-check -- --base=main
- touched-file ESLint: zero errors, with two pre-existing ComposerArea warnings
- full Vitest suite: 2,481 passed; one unrelated dogfood wrapper timeout passed 7/7 when rerun in isolation

### DIFF
--- a/src/app/preview/canvas-glass/use-send-buffer.tsx
+++ b/src/app/preview/canvas-glass/use-send-buffer.tsx
@@ -21,14 +21,19 @@
  * value lives in the composer).
  */
 
-import { useCallback, useEffect, useRef, useState } from 'react';
 import { motion } from 'framer-motion';
+import {
+  SEND_UNDO_GRACE_MS,
+  useComposerSendBuffer,
+  type ComposerSendBuffer,
+  type ComposerSendImage,
+  type QueuedComposerSend,
+} from '@/lib/hooks/use-composer-send-buffer';
 import { FONT } from './ui';
 
-export const SEND_UNDO_GRACE_MS = 4500;
-
-export interface ComposerImage { name: string; dataUri: string }
-export interface QueuedSend { id: number; text: string; images: ComposerImage[] }
+export { SEND_UNDO_GRACE_MS };
+export type ComposerImage = ComposerSendImage;
+export type QueuedSend = QueuedComposerSend;
 
 /** Where an undo would truncate the transcript back to. Returned by a
  *  successful dispatch; null means the send never went out (don't arm undo). */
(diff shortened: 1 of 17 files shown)


hurttlocker:
Merged with operator authorization after local verification. Every GitHub Actions job failed before runner allocation with the account billing/spending-limit annotation; no workflow steps ran. The merged commit had already passed local TypeScript, focused tests, rule-check, and touched-file ESLint, with the full suite passing except one unrelated timeout that passed 7/7 when rerun in isolation.

Issue #1621, opened 2026-07-27
cross-runtime model leak: stored codex defaultDispatchModel reaches claude-code worker spawn → instant model_not_found zero-diff failure


hurttlocker:
Live hit 2026-07-27 ~03:59Z dispatching #1620 with `runtime: 'claude-code'` (mission-ca6a3608-769, pkt-7fdbf5d6): the worker spawned with `--model gpt-5.6-sol`, Anthropic API returned 404 `model_not_found`, packet died in 26s with `zero_diff_failed`. Transcript: `~/.o8/owned-claude-code/claude-code-owned-1785124768399-a4ec0b09/runs/1785124768401-a0c5b9b1.jsonl` (final result block has `api_error_status: 404`, `--model gpt-5.6-sol` message).

Root cause shape: `resolveWorkerRouting` (src/lib/agents/routing.ts:117-123) already guards requested-model-vs-selected-runtime ("a Codex model hint can't leak onto a Gemini packet"). But the mission path also resolves the operator's stored `defaultDispatchModel` via `effectiveProfileModel` (src/lib/operator/subscription-profile.ts), which checks model-house against the SUBSCRIPTION PROFILE only — profile `both` allows any house — and never against the runtime actually selected. A codex id stored as the dispatch default therefore reaches non-codex spawns unfiltered.

Fix shape: apply the same house-vs-selected-runtime filter to the stored default (and any fallback source), not just `requestedModel` — fail to the runtime adapter's own `defaultModel` (claude-code: `MODEL_IDS.claudeWorkerDefault`) on mismatch. Per the reachability doctrine this needs a real-path test through the actual mission dispatch chain (create → dispatch → launch args) asserting a claude-code packet NEVER receives a codex model string — the routing.ts unit guard stayed green while the real path leaked.

hurttlocker:
Forensic update: clearing the operator defaultDispatchModel does NOT heal an existing mission — reset_packet + dispatch_mission relaunched with --model gpt-5.6-sol AGAIN (second transcript: ~/.o8/owned-claude-code/claude-code-owned-1785125999944-6ee954d9/runs/1785125999946-f76d4cdb.jsonl). The model is frozen into the mission/packet record at creation time and reused on redispatch. Fix must cover BOTH: (1) house-vs-runtime filter at model resolution, (2) redispatch re-resolving the model instead of replaying the frozen one (or the frozen value being validated against the runtime at launch). Real-path test should cover the reset→redispatch path too.

[...]

PR #1985, opened 2026-08-29
fix: expose incompatible model rejection

### LINKED ISSUE #1984: Explicit model mismatches silently fall back to the runtime default
## Problem

At current `main` (`29cecb558020395d489d9b9e18945ed0a9dac545`), `resolveWorkerRouting` correctly prevents a model from leaking into an incompatible runtime. When a model request does not target the selected runtime/provider/model house, however, it is reduced to `selectedModel: null`:

```ts
selectedModel: modelTargetsSelected ? requestedModel : null
```

The runtime then launches with its own default model. The routing receipt preserves `requestedModel`, but its `reason` only reports the selected runtime, so an explicit request can appear accepted even though a different model will run.

Example:

```ts
resolveWorkerRouting({
  requestedRuntime: 'claude-code',
  requestedModel: 'gpt-5.6-sol',
})
```

Today this selects `claude-code`, rejects the requested model to `null`, and silently delegates model choice to the Claude runtime default.

The house filter from #1621 is load-bearing and should remain. This issue is about making its rejection observable, not allowing cross-runtime model leakage.

## Expected behavior

An explicit incompatible model request should produce an observable, stable rejection disposition before launch. A runtime default remains valid when no model was explicitly requested.

The smallest useful slice could be either:

- a structured model-selection disposition on `WorkerRouting`; or
- another stable machine-readable rejection signal already native to the routing/packet contract.

The operator-facing reason should identify the requested model, why it was not selected, and that the runtime default will be used. If o8 prefers to fail the launch instead, the same test should prove the request fails closed rather than being silently substituted.

## Acceptance evidence

- A test first demonstrates the current mismatch: explicit incompatible model → `selectedModel: null` with no model-rejection explanation.
- The fixed routing result records an explicit incompatible-model disposition or error.
- No requested model continues to use the selected runtime's normal default without being labeled a rejection.
- Compatible exact model IDs and provider-qualified IDs for model-agnostic runtimes remain unchanged.
- A production entry-point test covers one mission/worker launch argument path so the disposition cannot exist only in the helper.
- The packet/audit receipt preserves requested model, selected model or runtime-default state, provider, runtime, and effort without inventing an identity.

## Non-goals

- Replacing runtime model catalogs or pickers.
- Adding fuzzy aliases or normalizing misspelled IDs.
- Removing the cross-house safety filter from #1621.
- Expanding which runtimes are dispatchable.
- Changing automatic runtime selection when the operator did not provide an explicit model.

I searched existing issues and PRs. #1621 fixed cross-runtime model leakage, #1834 fixed resolved-model persistence, and #1869 established that absent identity should remain unknown unless an upstream contract defines a default. I did not find an open issue or PR for making this explicit model-rejection path observable.

kvnloo:
> **Terminology correction:** The committed field is `modelDisposition`. The first bullet below retained the earlier working name `modelSelectionDisposition`; the implementation, tests, and remaining PR description refer to the three dispositions accurately.

> **Author-review gate:** This pull request remains draft. Despite the checked preparation-review box below, the submitting contributor must personally understand and be able to explain the complete diff before marking it ready; independent AI-assisted review does not replace that responsibility.

## What changed

- Add a stable `modelSelectionDisposition` to worker routing: `requested`, `runtime-default`, or `rejected-incompatible`.
- Explain incompatible explicit requests with the requested model, selected runtime, and the runtime-default fallback.
- Preserve and normalize the disposition in the durable worker routing receipt.
- Cover both the routing helper and the real mission/packet creation path.

## Why

An explicit model that does not belong to the selected runtime was correctly blocked from leaking across model houses, but the resulting `selectedModel: null` looked the same as an intentional runtime default. This makes that rejection observable without weakening the cross-runtime safety filter.

Closes #1984

## Verification

- [x] `npx tsc --noEmit`
- [ ] `npm test`
- [ ] `npm run lint`

Verified on Node v22.23.2:

- `npx vitest run src/lib/agents/routing-effort.test.ts tests/worker-model-routing-real-path.test.ts tests/single-sub-token-ladder.test.ts` — 3 files, 16 tests passed on the exact candidate against current `main`.
- `npx eslint src/lib/agents/routing.ts src/lib/agents/routing-effort.test.ts src/lib/orchestrator/types.ts tests/single-sub-token-ladder.test.ts tests/worker-model-routing-real-path.test.ts` — passed.
- `npm run rule-check -- --base=29cecb558020395d489d9b9e18945ed0a9dac545` — passed with zero violations on the two changed production files.
- `npm run test:classification:check` — passed.
- The full `npm test` suite reached 3,719 passed and 3 skipped tests, with one unrelated environment-sensitive `cli-locate` failure outside the changed paths, so the full-suite checkbox remains unchecked.

Regression proof: independently restoring the old null-only routing behavior made exactly two assertions fail in the helper and durable mission receipt tests (`runtime-default` versus `rejected-incompatible`). Restoring this change returned the focused four-file verification set to 20 passing tests.

## Risks and non-goals

- This adds one persisted routing field; old receipts are read as partial routing data and re-derived safely.
- It does not change runtime selection, model catalogs, fuzzy aliases, dispatchability, or the cross-house filter from #1621.
- It does not make an incompatible model launch under the selected runtime; that runtime still uses its default.

## Author review

This pull request was developed with AI assistance. I reviewed the complete diff, the routing and persistence behavior, and the verification evidence, and I am responsible for the change.

## Summary by CodeRabbit

* **New Features**
  * Added model routing status details showing whether a requested model was selected, rejected as incompatible, or replaced by a runtime default.
  * Improved routing explanations when a requested model is incompatible with the selected runtime.

* **Bug Fixes**
  * Prevented incompatible model requests from being treated as successfully selected.

* **Tests**
  * Added coverage for requested, defaulted, and rejected model-routing outcomes.

### DIFF
--- a/src/lib/agents/routing-effort.test.ts
+++ b/src/lib/agents/routing-effort.test.ts
@@ -18,6 +18,7 @@ describe('resolveWorkerRouting — effort parity (unset)', () => {
     // Untouched: default codex selection, no model.
     expect(r.selectedRuntime).toBe('codex');
     expect(r.selectedModel).toBeNull();
+    expect(r.modelDisposition).toBe('runtime-default');
   });
 
   it("'adaptive' is treated as no explicit effort (runtime default) ⇒ null", () => {
@@ -79,13 +80,19 @@ describe('resolveWorkerRouting — model house', () => {
     expect(routing.selectedRuntime).toBe('claude-code');
     expect(routing.requestedModel).toBe('gpt-5.6-sol');
     expect(routing.selectedModel).toBeNull();
+    expect(routing.modelDisposition).toBe('rejected-incompatible');
+    expect(routing.reason).toContain('gpt-5.6-sol');
+    expect(routing.reason).toContain('claude-code');
+    expect(routing.reason).toContain('runtime default will be used');
   });
 
   it('honors a Claude model hint for a selected Claude runtime', () => {
-    expect(resolveWorkerRouting({
+    const routing = resolveWorkerRouting({
       requestedRuntime: 'claude-code',
       requestedModel: 'claude-opus-5',
-    }).selectedModel).toBe('claude-opus-5');
+    });
+    expect(routing.selectedModel).toBe('claude-opus-5');
+    expect(routing.modelDisposition).toBe('requested');
   });
 
   it('rejects a Claude model for a selected Codex runtime', () => {
@@ -96,9 +103,11 @@ describe('resolveWorkerRouting — model house', () => {
   });
 
   it('preserves provider-qualified models for model-agnostic runtimes', () => {
-    expect(resolveWorkerRouting({
+    const routing = resolveWorkerRouting({
       requestedRuntime: 'opencode',
       requestedModel: 'google/gemini-2.5-flash',
-    }).selectedModel).toBe('google/gemini-2.5-flash');
+    });
+    expect(routing.selectedModel).toBe('google/gemini-2.5-flash');
+    expect(routing.modelDisposition).toBe('requested');
   });
 });
--- a/src/lib/agents/routing.ts
+++ b/src/lib/agents/routing.ts
@@ -140,12 +140,17 @@ function routingConfidence(intent: WorkerIntent): WorkerRoutingConfidence {
 }
 
 function routingReason(input: {
+  requestedModel: string | null;
+  modelTargetsSelected: boolean;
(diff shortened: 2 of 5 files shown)


hurttlocker:
Merged. Thanks for the clean disposition field and the real-path test through mission creation; that is exactly the shape we want. The red review-bot/label marks were fork-token artifacts, tracked in #1992.

PR #1986, opened 2026-08-29
fix: keep tmux CSI S out of dashboard scrollback

### LINKED ISSUE #1979: Terminal scrollback drops lines that tmux scrolls out with CSI S
## Symptom

The desktop terminal's scrollback silently loses lines under scrolling output. The visible screen is correct; scrolling up shows gaps (e.g. a 200-line numbered printf loop is missing lines 000, 005, 006, 010, 011, ...).

## Cause

The panel is attached to tmux with `TERM=xterm-256color`, which advertises `indn`. tmux (3.6a here) then scrolls the region with `CSI n S` instead of emitting line feeds. xterm.js implements `CSI S` by deleting the top `n` lines of the scroll region and inserting blanks at the bottom; those lines are never pushed into scrollback. Lines that scroll out via a plain LF at the bottom row *are* kept, which is why the loss is intermittent and batch-shaped.

Headless reproduction (`@xterm/headless` 5.5.0, 20×5, scrollback 1000):

- `L1..L5` + `L6\r\nL7` via LF → buffer `L1 L2 L3 L4 L5 L6 L7`
- `L1..L5` + `ESC[2S ESC[4;1H L6\r\nL7` → buffer `L3 L4 L5 L6 L7` (L1, L2 gone)

Found while building the terminal-workload correctness oracle for #1721: the rendered client buffer could never equal `tmux capture-pane -S -`, only the screen rows could.

## Fix candidates

1. Attach-client terminal override so tmux stops using `indn` for our client: e.g. `terminal-overrides ',xterm*:indn@'` on the attach (or an attach `TERM` whose terminfo lacks `indn`). tmux falls back to LF scrolling; xterm keeps every line. Cheapest; verify it doesn't slow busy panes.
2. Translate `CSI n S` in the panel's write path into `n` line feeds at the bottom of the region when the region is the full screen. Emulator-side, no tmux dependency, but must not fire for partial regions (TUIs).

Acceptance: the numbered-loop repro renders all 200 lines in scrollback in the real panel; a real-path test through the WebSocket + PTY path asserts the rendered buffer equals `capture-pane -S -` at quiescence.

kvnloo:
## What changed

- Override tmux's `indn` capability for dashboard attach clients so full-screen scrolling falls back to line feeds that xterm preserves in scrollback.
- Normalize captured tmux history to CRLF before replay so each row starts at column zero.
- Add focused persistence coverage plus a production WS → PTY → base64 → xterm real-path regression for the numbered-scrollback failure.

## Why

Dashboard terminals advertise `xterm-256color`, so tmux can emit `CSI n S` (`indn`) when output scrolls. xterm removes those rows from the scroll region without adding them to scrollback, producing the gaps reported in #1979. Disabling only `indn` at the tmux attach boundary preserves all lines while leaving partial-region TUI scrolling intact.

Closes #1979

## Verification

Node `v22.23.2`:

- [x] `npx tsc --noEmit`
- [ ] `npm test` — ran; final summary was 627 passed and 3 skipped, with one unrelated existing hermetic failure in `src/lib/runtimes/shared/cli-locate.test.ts`.
- [ ] `npm run lint` — repo-wide lint was not run; touched-file ESLint completed with 0 errors and 3 pre-existing unused-variable warnings in `src/ws-server.ts`.
- [x] `npx vitest run src/lib/ws-server/dash-terminal-persistence.test.ts tests/dash-terminal-csi-s-scrollback-real-path.test.ts` — 11 passed.
- [x] `npm run rule-check -- --base=29cecb558020395d489d9b9e18945ed0a9dac545`
- [x] Test-classification check.

Real-path evidence: the regression traverses the production WebSocket server, tmux-backed PTY, base64 terminal frames, and headless xterm buffer. The RED witness reproduced missing numbered lines while `CSI S` was present. GREEN retained the complete scrollback. As a sabotage check, removing the `indn@` override made the oracle fail as expected because `CSI S` returned. The partial-region TUI case still rendered marker `777`.

Hotspots: `src/ws-server.ts` and `tests/test-classification.json`; this PR intentionally limits both edits to the #1979 behavior and its real-path classification.

## Author review

## AI disclosure

This change was developed with AI assistance. I reviewed the complete diff, reproduced the failure and fix through the production path, ran the listed gates, and independently checked the sabotage and partial-region cases.

## Operator note

Dashboard attach now uses a dedicated tmux server (`-L o8-dashboard`). Sessions named `o8-dash-*` left on the default server by earlier builds are no longer found by `has-session` and will linger until the operator kills them once.

### DIFF
--- a/src/lib/ws-server/dash-terminal-persistence.test.ts
+++ b/src/lib/ws-server/dash-terminal-persistence.test.ts
@@ -78,14 +78,20 @@ describe('dashboard terminal persistence real fallback path', () => {
       execFileSync: exec,
       recordHealth,
     })).toBe(false);
-    expect(exec.mock.calls[3]?.[1]).toEqual(['kill-session', '-t', 'cortex-dash-test']);
+    expect(exec.mock.calls[3]?.[1]).toEqual([
+      '-L', 'o8-dashboard', 'kill-session', '-t', 'cortex-dash-test',
+    ]);
     expect(recordHealth).toHaveBeenCalledWith('degraded', 'session_create_failed');
   });
 
-  it('receipts a real backing session after all required options land', () => {
+  it('creates dashboard sessions on the isolated tmux server with only indn disabled', () => {
     const recordHealth = vi.fn();
     const exec = vi.fn()
       .mockImplementationOnce(() => { throw new Error('no existing session'); })
+      .mockReturnValueOnce(Buffer.from(''))
+      .mockReturnValueOnce(Buffer.from(''))
+      .mockReturnValueOnce(Buffer.from(''))
+      .mockReturnValueOnce(Buffer.from('xterm*:XT'))
       .mockReturnValue(Buffer.from(''));
     vi.spyOn(console, 'log').mockImplementation(() => {});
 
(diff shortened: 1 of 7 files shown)


hurttlocker:
Thanks for taking #1979 on; the CRLF replay fix for captured history is right and the real-path test is the correct shape. Two things block this as written:

1. `terminal-overrides` is a tmux **server** option. `set-option -ga -t <session>` does not scope it to the session: it appends to the global list for every session on that server (I verified on an isolated server: a second session sees the entry, and each create appends another duplicate, so the list grows by one every time a dashboard session is created or reused). Dashboard sessions live on the operator's default tmux server, so this also changes their personal tmux. It needs to be idempotent (check `show-options -g` before appending) and ideally scoped, e.g. a dedicated server (`-L`) for dashboard sessions, or a client-side approach.

2. `smcup@:rmcup@` disables alternate-screen entry/exit for every client, not just `indn`. Full-screen programs (editors, pagers, the alternate-screen phase of the terminal workload fixture) would leave their content in the primary buffer and never restore the previous screen. #1979 only needs `indn@`; if there is a reason to also drop the alternate screen, it should be its own change with its own test against the workload fixture's alternate-screen oracle.

Happy to re-review once it's `indn@` only, applied once per server, and the PR is marked ready.

kvnloo:
Maintainer repair (draft kept).

Head: `a30a9f4f40cb2ad843b649d6d82e5e3793c842a4` on current `main` (`ec6be55df`).

Scope:
- `terminal-overrides` is now `xterm*:indn@` only (no `smcup@`/`rmcup@`).
- Append is idempotent: `show-options -gv` then `-ga` only if `xterm*:indn@` is absent, so a second dashboard session on the same isolated tmux server does not grow duplicates.
- Accepted CRLF captured-history replay is unchanged.
- Dashboard tmux traffic stays on `-L` (env `O8_DASH_TMUX_SERVER_NAME`, default `o8-dashboard`).

Verified here:
- `npx vitest run src/lib/ws-server/dash-terminal-persistence.test.ts tests/dash-terminal-indn-isolated-server.test.ts` — 11/11 passed (mocked create/reuse + real isolated tmux server: one `indn@`, no smcup/rmcup, second session/reuse does not change overrides, alternate-screen oracle intact).
- `npx eslint` on the persistence sources — clean.
- Full WebSocket real-path file still present; this host did not get `/health` from `src/ws-server.ts` within 90s (better-sqlite3 self-register + bootstrap), so the isolated-server tmux oracle is the green proof for the maintainer contract.

No ready/merge.

hurttlocker:
CI is green on the rework (fork runs approved). The `indn@`-only override and the idempotent append are what was asked for.

One thing to cover before undrafting: moving dashboard sessions to a dedicated `-L o8-dashboard` server means sessions created by earlier builds on the default tmux server are no longer found by `has-session`, so they get recreated while the old ones linger on the default server until something kills them. Either a one-time sweep of `o8-dash-*` sessions on the default server at startup, or a short note in the PR body saying the operator should kill them once, would close that. Ready to merge on your flip after that.

PR #1987, opened 2026-08-29
fix: align orchestrator shortcut labels

kvnloo:
> **Author-review gate:** This pull request remains draft. The submitting contributor must understand and be able to explain the complete diff before marking it ready; independent AI-assisted review does not replace that responsibility.

## What changed

- Relabel Cmd/Ctrl+T as “New orchestrator tab” in the rendered shortcuts catalog and header tooltip.
- Align the dashboard shortcut comment with the production `spawnOrchestrator` action.
- Add a regression that traverses the exported rendered catalog and dispatches a real keyboard event through the production keydown handler.

## Why

The shortcut overlay described Cmd/Ctrl+T as “New tab,” but the production handler opens an Orchestrator tab. The label and bound action now agree, and the test pins that pairing against future drift.

Closes #1972

## Verification

- [x] `npx tsc --noEmit`
- [ ] `npm test`
- [ ] `npm run lint`
- [x] Other relevant checks are listed below, or skipped checks are explained.

Verified on Node v22.23.2:

- Focused regression: 2 files, 3 tests passed.
- RED: the new regression failed on the baseline because the rendered shortcut catalog was not exported for the production-pairing assertion.
- GREEN: the focused regression passed after the minimal label/catalog change.
- Sabotage: restoring the old “New tab” label produced the expected label mismatch; restoring the fix returned the focused suite to green.
- `npx eslint` on all four touched files passed.
- `npm run rule-check -- --base=29cecb558020395d489d9b9e18945ed0a9dac545` passed on the three changed production files.
- `npm run test:classification:check` and the repository diff check passed.
- The full `npm test` run reached 3,720 passed and 3 skipped tests, with one unrelated environment-sensitive `cli-locate` failure because the production scan found the installed `/usr/bin/gemini` before the test's injected fake nvm path. The full-suite checkbox therefore remains unchecked.
- Repo-wide `npm run lint` was not run; the required touched-file ESLint gate passed.

## Scope and risk

- Copy and test only; no styling, dependency, lockfile, persistence, authorization, or runtime-selection changes.
- The current-main terminal visibility changes from #1983 are retained; the dashboard page change is limited to the shortcut comment.

## Author review

- [x] I reviewed and understand the complete diff, including any AI-generated code.

This pull request was developed with AI assistance. I reviewed the complete diff and the verification evidence, and I am responsible for the change.

## Summary by CodeRabbit

* **Documentation**
  * Clarified that the ⌘T shortcut opens a new orchestrator tab.

* **UI Improvements**
[...]

PR #1988, opened 2026-08-29
fix: guard external release build cache

### LINKED ISSUE #1940: Release build cache: external-volume root with a mount guard and a size budget
Follow-up to the release intermediate cache (`scripts/lib/release-build-cache.mjs`).

## What happens

The cache root defaults to `~/.o8-build-cache/release-v1` on the boot volume. After two ships it holds ~10 GB (one entry per phase is kept, but each entry is several GB), on a machine that has wedged on ENOSPC before. `O8_RELEASE_BUILD_CACHE_DIR` already relocates it, and the operator wants it on the external archive SSD.

## Expected

1. A ship-time guard in the ship wrapper: when `O8_RELEASE_BUILD_CACHE_DIR` (or an operator setting) points under `/Volumes/<drive>`, use it only if the volume is actually mounted — check for a known marker file at the volume root, not just the directory — otherwise fall back to the local default and log one line saying so. Writing under `/Volumes/<name>` while the drive is unmounted must never happen (macOS would create a boot-disk folder that later shadows the real mount).
2. A size budget for the cache root (default ~4 GB when local, larger when external), pruning oldest entries beyond it, logged.
3. Docs: `docs/operations/release-build-cache.md` gains the external-volume recipe and the marker-file rule.

## Proof

Hermetic tests for the guard (marker present → external root; marker absent → local root + log line) and for the budget prune (entries beyond the budget removed oldest-first, receipts untouched). A dry-run ship on this machine with the cache root pointed at the external drive.

kvnloo:
## What changed

- Guard the configured external release-cache root with a non-symlinked mount marker whose recorded identity must match the live mount.
- Fall back to the local cache without spilling writes onto the boot volume when the external mount disappears, drifts, or fails path-safety checks.
- Add deterministic byte-budget pruning that preserves receipts and fails closed when cache entries escape the root through symlinks.
- Document the resolver, capture, mount-marker, and budget behavior.

Closes #1940

## Why

A configured `/Volumes` path can remain present as an ordinary directory after its external volume disappears. Release capture must not silently write large artifacts to the boot disk, and cache growth needs a deterministic bound without risking receipts or following path escapes.

The focused regression was RED on the approved baseline: 3 tests failed because the production resolver accepted an unguarded configured root and budget enforcement/pruning were absent. GREEN exercises the production resolver and `captureReleaseBuildCache`: matching identity uses the external root; disappearance, identity drift, symlinked markers, and escaped entries fail closed; capture falls back locally without external spill; pruning is oldest-first with deterministic ordering and preserves receipts.

## Verification

- [x] `npx tsc --noEmit`
- [ ] `npm test`
- [ ] `npm run lint`

Verified on Node `v22.23.2`:

- `npx vitest run tests/release-build-cache.test.ts` — 1 file, 15 tests passed.
- `npx tsc --noEmit` — passed.
- `npx eslint scripts/lib/release-build-cache.mjs scripts/lib/release-build-cache.d.mts tests/release-build-cache.test.ts` — passed.
- `npm run rule-check -- --base=29cecb558020395d489d9b9e18945ed0a9dac545` — passed with zero violations.
- `git diff --check 29cecb558020395d489d9b9e18945ed0a9dac545..HEAD` — passed.
- Repo-wide `npm test` and `npm run lint` were not run; the focused behavioral suite, mandatory TypeScript gate, touched-file ESLint, and changed-line rule check passed.

## Scope and residual

- The branch contains one commit and exactly four changed files: the resolver/capture implementation, its declaration, focused tests, and operations documentation.
- There is no changed-path overlap with open PRs #1985, #1986, or #1987.
- Linux cannot perform a physical macOS `/Volumes` hardware dry-run. Production resolver/capture behavior for mount disappearance, identity drift, symlink/path escape, boot-volume no-spill, deterministic pruning, and receipt preservation is directly exercised.

## Author review

This pull request was developed with AI assistance. The complete diff and independent approval evidence were reviewed, and I am responsible for the change.

### DIFF
--- a/docs/operations/release-build-cache.md
+++ b/docs/operations/release-build-cache.md
@@ -31,6 +31,18 @@ This verification detects corruption and metadata mismatch. It does not turn an 
 
 Set `O8_RELEASE_BUILD_CACHE_DIR` to isolate a canary or move the cache to another local volume. Set `O8_RELEASE_BUILD_CACHE=off` to bypass restore and capture without changing the build command.
 
+For a macOS external volume, create a private marker at the volume root whose contents are a stable operator-chosen identity, then configure both the cache directory and that identity:
+
+```bash
+export O8_RELEASE_BUILD_CACHE_VOLUME_ID="<stable-volume-identity>"
+printf '%s\n' "$O8_RELEASE_BUILD_CACHE_VOLUME_ID" > "/Volumes/<volume>/.o8-release-cache-volume"
+export O8_RELEASE_BUILD_CACHE_DIR="/Volumes/<volume>/o8/release-cache"
+```
+
+Create the marker while the intended volume is mounted. Do not pre-create `/Volumes/<volume>` when it is absent. Before every restore or capture, the wrapper reads the marker and requires its contents to exactly match `O8_RELEASE_BUILD_CACHE_VOLUME_ID`. A missing marker, missing identity, or changed marker causes one warning and selects the local default instead; the wrapper never creates the configured `/Volumes/...` path in that state. `O8_RELEASE_BUILD_CACHE_LOCAL_DIR` can override the fallback for a hermetic canary.
+
+The cache-entry budget defaults to 4 GiB on the local root and 20 GiB on a validated external root. Set `O8_RELEASE_BUILD_CACHE_MAX_BYTES` to a non-negative integer byte count to override it. Capture prunes whole compatibility entries oldest-first, with relative path as the deterministic tie-breaker. Receipts are outside the budget and are not pruned.
+
 The cache retains one source entry for each compatibility identity and the two newest compatibility identities for each phase. This bounds the large native and web intermediates while preserving the most useful rollback point.
 
 ## Receipts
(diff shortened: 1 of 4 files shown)


hurttlocker:
Read through; the change is scoped and the tests are the right shape. Two mechanical notes before it can merge: mark the PR ready when your author-review gate is satisfied, and it needs a rebase onto current `main` so CI runs against the tree it will land on. I approve the fork workflow runs after reading each diff, so once it's ready and green I'll merge.

hurttlocker:
Current checkpoint: this fix is directly relevant. The local release cache has reached 9.8 GB, and current main still lacks the external-volume guard and deterministic byte budget. The direction of this diff remains right. Please mark the PR ready and rebase it onto current main when you can; I will rerun the Mac gates and merge it if clean. If you cannot pick it back up, say so and I will adopt the patch with attribution rather than leave the storage risk open.

hurttlocker:
Adopted onto current `main` in #2069, with this pull request preserved as the source and attribution record. Independent review added a top-level `entries` symlink guard before budget pruning. The complete cache-budget fix is now landed, so I am closing this superseded draft.

PR #1991, opened 2026-08-29
Probe OpenAI-compatible local model endpoints

### LINKED ISSUE #1944: Probe configured local endpoints through both supported model-list protocols
Child of #1451. Picked from the gap list in `docs/internals/local-model-coverage.md`.

## Problem
Local-model discovery probes only `/api/tags` (the Ollama-style model list; see `src/lib/cortex/qa/llm/inference-route.ts` and `src/app/api/setup/detect/route.ts`). Local servers that speak the OpenAI-style `/v1/models` list — LM Studio, vLLM, llama.cpp's server — answer nothing to `/api/tags`, so a configured endpoint that works for inference shows no models in Settings and reads as dead in Diagnostics.

## Expected
The liveness/model-list probe tries `/api/tags` and, when that fails, `/v1/models`, normalizing both into the same model list shape. The resolver caches the result as it does today; a server that answers either protocol counts as live.

## Proof
Unit tests for the probe with mocked fetch: tags-only server, models-only server, neither (dead), both (tags preferred). A component test that Settings lists a model returned only by `/v1/models`.

techlowneko:
## What changed

- Probe configured local inference endpoints through `/api/tags` first, then fall back to `/v1/models` when the Ollama protocol fails or returns an unrecognized payload.
- Normalize Ollama model names and OpenAI-compatible `data[].id` values into the existing `models: string[]` result while preserving URL normalization and caching.
- Cover tags-only, models-only, neither-protocol, tags-preferred, transport-failure fallback, unrecognized-schema fallback, valid empty lists, cache, and Settings `running / 1 models` behavior.
- Update the local-model coverage documentation to describe both protocols and remove this completed work from the open-gap list.

## Why

OpenAI-compatible local servers such as LM Studio can be healthy without implementing Ollama's `/api/tags` endpoint. The shared probe previously reported those configured endpoints as offline.

Closes #1944.

## Verification

- [x] `npx tsc --noEmit`
- [ ] `npm test` — the equivalent full Vitest command was run directly under pinned Node 22 after rebuilding native dependencies for that ABI. It completed with 584 files / 3,660 tests passing and 47 files / 67 tests failing in unrelated Windows/platform areas (POSIX commands and paths, symlink/permission behavior, and Windows file-lock cleanup). Both changed test files pass.
- [x] `npm run lint` — the equivalent repository-wide ESLint command passed under pinned Node 22 with 0 errors and the configured maximum of 152 existing warnings.
  - Focused Vitest: 2 files, 29 tests passed under Node 22.
  - Live Ollama probe: running with six discovered models.
  - Live LM Studio probe: both model-list protocols reachable.
  - Live fallback integration: a loopback server returned 404 for `/api/tags` and forwarded the installed LM Studio server's real `/v1/models` response; the production probe requested both paths in order and reported a running endpoint with discovered models.
  - `npm run rule-check -- --base=upstream/main` passed.
  - `git diff --check` passed (with expected Windows line-ending warnings only).

## Author review


hurttlocker:
Merged. The /api/tags → /v1/models fallback with the parser split reads well, and it closed #1944. The red review-bot mark was a fork-token artifact, tracked in #1992.

PR #2009, opened 2026-08-30
Fix dev NODE_OPTIONS stub path resolution

### LINKED ISSUE #1994: Dev-mode NODE_OPTIONS uses a relative server-only stub path that breaks child Node processes outside the repo root
## Summary

`scripts/dev.mjs` (and the `bench:*` / `smoke:*` scripts) set `NODE_OPTIONS=--import=./scripts/register-server-only-stub.mjs` with a path relative to the repo root. `NODE_OPTIONS` is inherited by every child Node process the dev servers spawn; any child whose working directory is not the repo root fails at startup with `ERR_MODULE_NOT_FOUND: Cannot find module '<cwd>/scripts/register-server-only-stub.mjs'`. Observed three times in one `npm run dev` boot with `<cwd>` = `~/.o8/worktrees` (children spawned by the realtime server / codebase-memory boot pass), and it would hit any Node-based tool a dispatched worker runs in dev mode (including the `o8` CLI).

## Direction

Resolve the stub path to an absolute file URL at script start (`pathToFileURL(new URL('./register-server-only-stub.mjs', import.meta.url))`) before exporting `NODE_OPTIONS`, in `scripts/run.mjs` / `scripts/dev.mjs` and the `package.json` scripts that inline it. One test: spawn `node -e ""` with `cwd: os.tmpdir()` under the exported env and assert it exits 0.

techlowneko:
## Summary

- resolve the inherited `server-only` stub import to a module-anchored absolute file URL
- reuse the shared helper across dev, shell-neutral smoke/bench launchers, memory bench, and Symon verification while preserving unrelated `NODE_OPTIONS`
- add a temp-directory real-runner regression that proves the stub is inherited and executed

## Verification

- `node --version` — `v22.22.0`
- focused integration test — 2 passed
- test-classification manifest check — passed
- `npx tsc --noEmit` — passed
- `npm run typecheck` — passed
- touched-file ESLint — passed
- `npm run lint` — passed (0 errors; 152 existing warnings)
- `npm run rule-check -- --base=upstream/main` — passed
- `git diff --check` — passed

## Local Windows limitations

- `npm test`: 3,782 passed; 69 unrelated platform/environment failures involving POSIX assumptions, Windows symlink privileges, and locked SQLite cleanup. The focused regression passes.
- `npm run protocol:check`: the raw check differs because of CRLF checkout normalization; the LF-normalized contract hash exactly matches the committed generated hash.

Closes #1994

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of server-only Node runtime options across development, benchmarking, verification, and command-running workflows.
  * Automatically normalizes legacy or duplicate options to ensure consistent child-process behavior.

* **Tests**
  * Added coverage for option normalization and child-process execution using legacy configuration.

### DIFF
--- a/scripts/bench/run-memory.mjs
+++ b/scripts/bench/run-memory.mjs
@@ -2,6 +2,7 @@
 import { spawnSync } from 'node:child_process';
 import fs from 'node:fs';
 import path from 'node:path';
+import { withServerOnlyStubNodeOptions } from '../run-lib.mjs';
 
 const CATEGORIES = [
   'ownership',
@@ -117,7 +118,7 @@ function main() {
       ...process.env,
       O8_EVAL_MODE: '1',
       THREE_WAY_LIMIT: process.env.BENCH_QA_LIMIT ?? undefined,
-      NODE_OPTIONS: '--import=./scripts/register-server-only-stub.mjs',
+      NODE_OPTIONS: withServerOnlyStubNodeOptions(),
     },
   });
 
--- a/scripts/dev.mjs
+++ b/scripts/dev.mjs
@@ -3,7 +3,7 @@ import { existsSync, mkdirSync, readFileSync, rmSync, writeFileSync } from 'node
 import { homedir } from 'node:os';
 import { join } from 'node:path';
 import { spawn } from 'node:child_process';
-import { resolveSpawn } from './run-lib.mjs';
+import { resolveSpawn, withServerOnlyStubNodeOptions } from './run-lib.mjs';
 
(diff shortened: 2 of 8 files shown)


hurttlocker:
Merged. The module-anchored import plus the canonicalizing helper is the right shape, and the regression running the real script runner from a temp directory is exactly the test the issue asked for. Thanks for taking it from the issue to a landed fix in one pass, and for spelling out the Windows-only suite failures separately.

PR #2114, opened 2026-09-08
fix: reduce populated-profile idle work

hurttlocker:
## Summary

Reduce background work exposed by the populated-profile hidden-idle check in 0.1.741. Prepare the synchronized 0.1.742 manifests; this PR does not itself publish an installer.

- Read lane histories only for packets in the mission being reconciled. Retain all matching attempts, ordering, fresh runtime checks and persisted decisions.
- Run worktree maintenance for active work, releases, terminal transitions and storage-held retries. Dormant queued records still reconcile but no longer trigger repeated filesystem sweeps.
- Send a review-only visibility hint from the desktop. Hidden clients keep durable inbox and transcript subscriptions. Reopening refreshes the full review view; another visible or legacy client still requests scans.
- Give two fake-runtime retry fixtures explicit stop receipts. Production stop confirmation is unchanged. Both fixture failures were also reproduced on the unchanged release baseline.

## Verification

- Full hermetic suite: 4,170 passed, three existing skips.
- Focused mounted-provider and persisted headless tests: 21 passed.
- Relevant resource-owning integration tests: 40 passed, including the real WebSocket server, real Git edits, hidden/reopen delivery, multi-client review demand, retry and read-only enforcement.
- TypeScript, touched-file lint, test classification, changed-line rules and whitespace checks passed.
- A private snapshot comparison of 36 mission states found no decision differences after scoping lane reads. Median reconciliation time was about 987 ms before and 52 ms after; this is a function-level measurement, not a whole-app performance claim.
- No visual layout change. Native populated-profile acceptance remains pending; the failed installed 0.1.741 receipt is preserved.

Refs #1817, #1697. Neither broad issue is closed by this source change. Runtime defaults, credentials, live packet state and retained worktrees are unchanged. Other operating-system and stronger-isolation work remains deferred.

PR #2118, opened 2026-09-09
fix: preserve fleet session identity and activity times

### LINKED ISSUE #2115: cortex_fleet_status lists the IDE-owned orchestrator session twice and renders lastActive as "NaNd ago" (0.1.740)
## Mechanic

On 0.1.740, a fresh Solo orchestrator session in the desktop app called `cortex_fleet_status` (the canned "Review pending changes" home-screen prompt does this) and the tool returned **two entries for the one live session**, both `status: reviewing`, both with `lastActive: NaNd ago`. The orchestrator itself flagged it in its reply as a duplicate registration plus a stale-timestamp bug, and had to explain it away to the operator before answering the actual question.

Two defects in one surface:

1. **Duplicate registration.** The IDE-owned orchestrator session appears twice in the fleet listing. Expected: exactly one row per live session key.
2. **`NaNd ago`.** The relative-age formatter receives a timestamp it cannot parse and emits `NaN`. This is the same symptom as #1859 (closed 2026-08-25, `relativeAge` guard for `o8_status`), now showing through the `cortex_fleet_status` path, so either the guard did not cover this call site or the timestamp source regressed.

## Repro

1. Desktop app 0.1.740, any repo, empty home screen, Solo mode.
2. Click the "Review pending changes" chip (or ask the orchestrator to list fleet status).
3. Read the `cortex_fleet_status` result in the tool call: the current session is listed twice, `lastActive: NaNd ago`.

Seen while recording a product demo; the raw tool output was on screen in the session transcript.

## Acceptance

- `cortex_fleet_status` returns one entry per live session key; a test registers the same IDE session twice (re-registration on reconnect) and asserts a single row.
- The relative-age formatter used by `cortex_fleet_status` never returns `NaN`: a test feeds it a missing timestamp, an ISO string, and an epoch number, and asserts a finite duration or an explicit "unknown" label. If #1859's guard lives in a different helper, share one formatter and delete the other.
- A just-launched session shows an age in seconds, not days.

hurttlocker:
Reconnecting an IDE-owned session could make `cortex_fleet_status` report the same session twice. Owned discovery also parsed relative labels such as `just now` as dates, producing `NaNd ago`.

Preserve qualified owned-session IDs when reading saved tabs. Carry the activity instant separately from its display label, and share one guarded age formatter between owned fleets and runtime inventory. Recent activity displays in seconds; missing or invalid timestamps display `unknown`. Distinct sessions with matching labels remain separate.

Fixes #2115.

Validation:

- Reproduced duplicate rows and invalid ages on the unchanged baseline before applying the fix.
- 34 focused checks passed, including the source MCP server calling the real inventory route over HTTP with isolated persisted session and tab fixtures. The fixture verifies repeated registration yields one row and preserves the saved activity instant without rewriting session state.
- Full hermetic suite: 4,167 passed, three existing skips.
- TypeScript, touched-file ESLint, changed-line rules, test classification and whitespace checks passed.

Packaged-app verification remains a release check.


hurttlocker (commented):
Direct review of 432ec573a6d2dbb6837c511e0e2b05a07245a0da found no blocking issue. The qualified owned-session identifier is now preserved, and the activity timestamp reaches discovery and inventory without reparsing a display label. Independently reran 28 focused checks, including the real MCP-to-HTTP inventory path, invalid timestamps, duplicate persisted registrations, and distinct same-label sessions. Merge remains gated on the updated branch checks after incorporating #2114. Packaged desktop verification is still a release check.

PR #2119, opened 2026-09-09
fix: preserve ACP reply chunks in thread rendering

### LINKED ISSUE #2117: ACP-backed orchestrator replies render with hard line breaks mid-sentence in the thread view (0.1.740)
## Mechanic

When the orchestrator is an ACP-backed runtime (the OpenCode 2 backend, any of its models), streamed reply text lands in the thread view with a hard line break every few words, so a one-sentence answer renders as five short lines:

```
I'll read index.html first, then answer in one sentence with
no edits.
I'd persist the list in localStorage first — added
tasks and checkbox state currently
die on every reload.
```

The same reply from a native Claude or Codex orchestrator renders as one paragraph. Markdown in these replies also breaks the same way (`**Techn` / `ical change:**` split across lines was observed with a different model on the same backend), which points at the streaming chunk boundaries being committed as line breaks rather than at the model's own newlines.

## Repro

1. Desktop app 0.1.740, OpenCode 2 runtime enabled in settings, any model from its catalogue.
2. In an orchestrator thread, hand off to the OpenCode 2 model (Hand off card) and ask a one-sentence question.
3. Watch the reply stream: chunk boundaries appear as line breaks and remain after the turn completes.

Seen while recording a product demo; two different models on the backend showed it, so it is the path, not the model.

## Acceptance

- Streamed text from an ACP backend is appended without inserting newlines at chunk boundaries; a test feeds a sentence split into arbitrary chunks and asserts the rendered text equals the joined sentence.
- Markdown emphasis spanning chunk boundaries renders correctly after the turn completes.
- Native and ACP orchestrators produce identical rendering for identical text.

hurttlocker:
ACP reply chunks were joined with newlines in the thread view, splitting sentences and breaking Markdown emphasis across chunk boundaries. Apply the existing verbatim text assembly rule to all ACP orchestrator backends, preserving both reply and reasoning text, including intentional newlines.

Fixes #2117.

Validation:

- Nine regression cases failed before the fix and pass afterward. They exercise the ACP event mapper, real React thread hook, and Markdown renderer with character-sized chunks, completion before the queued frame, tool-first output, split emphasis and inline code, explicit newlines, and native/backend rendering parity.
- Focused stream/client coverage: 58 tests passed.
- Full hermetic suite: 4,164 passed, 3 skipped.
- TypeScript, touched-file ESLint, rule-check, test classification, and whitespace checks passed.

Rendering verification uses a DOM fixture with mocked transport. The installed desktop app has not been rebuilt or visually rechecked for this patch.


hurttlocker (commented):
Direct review of 71914ddbaebc31437d0b97a845a5b1d75c26ccf9 found no blocking issue. The ACP-backed orchestrator paths emit text deltas, so applying the existing verbatim assembly rule preserves spaces, intentional newlines, emphasis and inline code rather than inserting a newline at each chunk boundary. Independently reran all 58 focused hook, socket and ACP client tests, including the actual React hook and Markdown rendering fixture. This does not claim installed-app visual verification. Merge will follow a normal current-main update and its required checks.

Issue #1817, opened 2026-08-22
Footprint: o8 should be light on a user's machine, measured not assumed


hurttlocker:
## Goal
o8 runs all day on an operator's machine next to their editor, browser, and (for us) a stream encoder. It must be small and quiet. Today we have no number for what it costs, and we have already found it burning CPU on background probes (#1806) and stalling dispatch on an 8s `du` (#1811).

## Scope
- **Measure first**: a repeatable footprint report — idle RSS and CPU for the Tauri shell, the bundled Next server, and ws-server; the same with N lanes running; disk under `~/.o8` over a week; and the count of child processes spawned per minute at idle.
- **Budget**: publish target numbers in the repo (idle CPU, idle RSS, processes/min) and fail a check when a build regresses past them.
- **Known offenders to fold in**: per-poll `lsof` (#1806), `du` on the dispatch path (#1811), the WKWebView leak history, poll intervals that run when nothing is on screen, and any timer that fires while the window is hidden.
- **Idle means idle**: with no lanes and the window hidden, background work should approach zero — no repeated process probes, no filesystem walks.

## Real-path test
A measurement harness that boots the app headless, records the metrics above at idle and under load, and asserts them against the budget; run it in the ship path so a regression is caught before release.

hurttlocker:
First measured numbers for this issue. **These are a contended-machine reading, not a baseline** — see the caveat section before drawing conclusions from the CPU figures. An idle reading is planned for comparison, and the commands are recorded below so the two are directly comparable.

Measured 2026-08-25 ~12:20 EDT, o8 v0.1.711, macOS, 8 physical / 16 logical cores, 64 GB.

## Download — no concern

| Asset | Size |
|---|---|
| `o8_0.1.711_x64.dmg` (first install) | **88 MB** |
| `o8.app.tar.gz` (auto-update) | **59 MB** |

Unremarkable for a desktop app.

## Installed — 213 MB

```
/Applications/o8.app          213 MB
  Contents/Resources          154 MB   bundled Next server + node bundles
  Contents/MacOS               58 MB   Rust binary
```

## Running — 8 processes, ~1.5 GB resident

| Process | RSS | Uptime |
|---|---|---|
| `o8` (Tauri shell + WKWebView) | **870 MB** | 15h 29m |
| `ws-server.mjs` | **357 MB** | 15h 29m |
| operator MCP server | 175 MB | 3d 16h |
[...]

Issue #1697, opened 2026-08-03
Add interaction-level performance budgets for boot, typing, navigation, and scale


hurttlocker:
## Problem

`npm run bench:speed` measures dashboard HTTP TTFB, bootstrap API time, CLI status, MCP overhead, and socket count. Those are useful service metrics, but they do not prove that the desktop interaction loop stays fast under the workloads o8 creates.

There is no repeatable gate for full app readiness, terminal keystroke-to-paint latency, tab switching, active-workspace reveal in a large fleet, Design Mode selection, main-thread stalls, or long-session resource growth. Performance work can therefore close after a sprint without leaving a regression detector for the operator-visible paths.

## Harness contract

Build deterministic fixtures and runners for these scenarios:

- Cold desktop launch to shell visible, workspace restored, and first interactive command accepted.
- Warm relaunch with persisted terminal, agent, canvas, and browser state.
- Terminal keystroke-to-paint latency with 1, 4, and 12 live terminal surfaces.
- Tab and pane switching while transcripts and diffs stream.
- Repo/worktree/agent navigation with generated fleets at 50, 250, and 1,000 rows, including active-context reveal.
- Design Mode arm, hover, selection, screenshot crop, and prompt-ready latency on a local fixture page.
- Idle memory, process count, WebSocket count, and main-thread long tasks during a bounded soak.

Each result must include commit SHA, app version, build mode, hardware/OS class, fixture size, sample count, p50/p95/p99 where meaningful, and raw artifact paths.

## Budgets and gating

Run a baseline against two current release builds before choosing final absolute budgets. Check the budget manifest into the repo with the reasoning for each threshold.

Use both absolute and relative gates: a result fails when it breaks the user-facing budget or regresses materially against the accepted baseline. Machine-sensitive metrics need a normalized local lane or dedicated runner rather than noisy generic CI.

Keep quick smoke measurements separate from long soaks. The normal completion gate must stay bounded, while scheduled/manual runs can cover 8-hour memory behavior and high-cardinality fleets.

## Acceptance tests

- [ ] One command generates fixtures, runs the quick suite, and writes stable machine-readable results.
- [ ] A deliberately injected render delay fails the relevant interaction budget.
- [ ] Results distinguish server wait, main-thread work, React commit, and paint where the platform exposes them.
- [ ] Failed or unavailable measurements are explicit nulls with reasons, never silent passes.
- [ ] Baseline comparison reports both absolute values and deltas.
[...]

PR #2162, opened 2026-09-11
test: avoid full scrollback scans during terminal marker polling

hurttlocker:
## Summary

The shared terminal benchmark marker poll requested a full snapshot every 250ms.
Full snapshots join every retained scrollback and rescan alternate-screen state,
even though the polling caller only needs bounded output tails. Profiling showed
this observer work consuming measurable realtime-server CPU.

- Add an explicitly identified output-tail response for marker polling.
- Keep full snapshots, retained escape-state checks, delivery behavior, polling
  frequency, and all numerical budgets unchanged.
- Reject servers that do not acknowledge the lightweight protocol rather than
  silently mixing measurement contracts. Record both request counts and contract v3.

## Verification

- 14 focused tests passed, including all existing budget checks.
- Tests prove output-only polling never inspects retained scrollback, full snapshots
  still follow alternate-screen retention changes, tails stay bounded, and responses
  cannot mutate the internal stats store.
- TypeScript, touched-file lint, whitespace and changed-line rules passed.
- Full hermetic suite and real 1/4/12-terminal verification are being completed;
  their results will be added before merging. No performance pass is claimed yet.

The preceding quieter v2 run passed every latency/correctness/memory limit, but
still missed renderer CPU p95 (35.85% vs 35%) and realtime CPU median (25.81%,
required below 25%). Those failed receipts are retained; this change does not
erase them or claim a product speedup from less benchmark work.

Related: #1697 and #1817. Neither broader issue is closed by this change.
No version bump, publication, native-app replacement, or changes to operator data.

### DIFF
--- a/docs/internals/terminal-workload-baseline.md
+++ b/docs/internals/terminal-workload-baseline.md
@@ -100,6 +100,15 @@ profiles beside the raw samples. These runs include profiler overhead and are
 explicitly rejected by the budget checker, even if their numbers are below the
 limits. An ordinary, unprofiled run is required for acceptance.
 
+Measurement contract v3 also separates marker polling from full correctness
+snapshots. The shared 250 ms marker poll reads only bounded output tails; it no
+longer joins and scans every terminal's retained history during each tick. Full
+snapshots still inspect retained alternate-screen sequences before the correctness
+assertions. The response schemas distinguish these paths, and the runner refuses
+a server that does not honor the lightweight request. Both request counts remain
+in each sample. This removes observer work, not product work, and does not change
+any budget or correctness requirement.
+
 Terminal-grid setup waits are bounded at 30 seconds and excluded from every measured latency.
 Visible-input polling reads 1,000 terminal lines. A timeout records whether the
 marker reached the auxiliary stream, the panel write path, and a painted xterm
--- a/docs/internals/terminal-workload-baseline.md
+++ b/docs/internals/terminal-workload-baseline.md
@@ -109,6 +109,13 @@ a server that does not honor the lightweight request. Both request counts remain
 in each sample. This removes observer work, not product work, and does not change
 any budget or correctness requirement.
 
+Contract v4 starts the browser counters after pre-workload memory probes and
+captures them before post-workload memory probes. Browser frame and long-task
+rates use the browser counters' own clock interval, not a shorter runner interval.
+Earlier receipts could report inflated rates, even more than the display cadence,
+because slow memory inspection extended the counter window without extending its
+denominator. Raw older results remain historical evidence, not comparable v4 rates.
+
 Terminal-grid setup waits are bounded at 30 seconds and excluded from every measured latency.
 Visible-input polling reads 1,000 terminal lines. A timeout records whether the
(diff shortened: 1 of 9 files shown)


hurttlocker:
Local completion gates are now green: 4,298 hermetic tests passed with four skips; the 14-test focused group passed; explicit TypeScript, touched lint, classification, and changed-line checks passed. The real 1/4/12-terminal workload is running against this exact source head with the production frontend. It will verify the new output-only response through the actual realtime entry point while retaining the full screen/escape-state assertions. No numerical budget or product terminal behavior was changed.

hurttlocker:
The first CI run passed all 4,298 assertions but correctly failed on two unhandled teardown errors. Its retained logs identify tests/ui-edit-runtime-preset-route.test.ts: asynchronous terminal-manifest and reservation cleanup ran after that routing-only fixture deleted its data directory. The terminal snapshot tests passed. I am correcting the fixture isolation before another CI run; no error suppression or gate bypass will be used.

hurttlocker:
Two follow-up corrections are now pushed. The routing-only fixture mocks terminal cleanup boundaries and asserts those hooks are invoked; real manifest lifecycle/registry tests passed separately. Browser counters also now start after pre-workload memory probes and are captured before post-workload probes. Frame/long-task rates use the browser clock that owns the counters. The previous runner could report above-display frame rates by dividing a longer counter window by the shorter workload interval. Contract v4 records the corrected interval; old raw failures remain retained. Focused browser-window/budget tests and explicit TypeScript passed. Updated full local and remote suites are running.

hurttlocker:
Reviewed the nine-file change and verified the real v4 workload: nine samples at 1/4/12 terminals, all input markers observed, all screen-correctness and rapid-switch checks intact. The corrected browser window reports roughly 60 frames/s. Numeric acceptance is still FAIL on two CPU rows (renderer p95 36.81% over 35%; realtime median 25.91% must be below 25%). These remain performance follow-up findings, not hidden or relaxed gates. This PR fixes observer cost, timing, and a routing-fixture teardown race; it does not claim the product meets every performance limit. Local suite: 4,300 passed, four skipped. All required CI checks passed on current merged-base head 309b8a2dded733547c3807f31a9ecc4b084c7361. No release publication or daily-app replacement.

hurttlocker:
Acceptance decision after the final measurement: the maintainer requested a bounded CPU tolerance and moving on from these near-target results. For this terminal pass, a 10% relative margin was selected and disclosed for only the two original failing counters: renderer CPU p95 36.81% is within 38.5% (original target 35%), and realtime CPU median 25.91% is below 27.5% (original target below 25%). Every other locked check passed unchanged. All 27 input markers were observed with zero timeouts; N=12 input p95 was 67.6ms and first correct frame p95 was 371.9ms.

This pass is accepted under that scoped tolerance. The original checker, raw measurements and two stricter-budget failures remain unchanged. This is not a CPU speedup claim, a global budget change, a release-baseline acceptance, or a native-renderer parity claim. No further CPU optimization is required for this pass unless real use regresses or measurements exceed the bounded margin. No release was performed. The separate later interaction work remains in #2163.

PR #2170, opened 2026-09-11
fix: bound completed dependency download caches

hurttlocker:
## Change

Bound the download caches created by `runDependencyInstall`. Completed recipe caches use a 2 GiB allocated-payload limit, a 12-entry limit, and 14-day expiry, evicting the least recently used eligible recipe first. Maintenance runs after installs, not on an idle timer or a dispatch preflight walk.

New writers use a separate managed namespace. Existing caches, installed dependency views, held workspaces, and conversation history are preserved. This contributes the missing package-cache retention mechanism to #1817; it does not close the installed-app footprint issue by itself.

## Safety

- Install reservations and eviction share the existing cross-process lifecycle lock. The managed root is bound to that lock database, including custom cache roots.
- Concurrent users of one recipe retain the cache until the last successful runner finishes. Failed or interrupted runners retain their reservations because parent exit alone cannot prove child processes stopped.
- Eviction requires recorded directory identity and size. Changed identities, symlinks, invalid metadata, and interrupted retirements remain held.
- Eligible namespaces are detached before the existing exact-directory purge. Installed dependency views remain valid after download-cache eviction.
- A persisted receipt distinguishes `within-budget` from `held`, makes unknown sizes explicit, and limits the accounting claim to the new managed namespace. Limits are initial policy defaults, not a whole-profile storage promise.

## Verification

- Hermetic suite: 4,300 passed, four skipped.
- Existing install coverage plus the initial retention suite: 44 passed.
- Final retention, repository-setup and saved-recipe route coverage: 16 passed.
- TypeScript, touched-file ESLint, classification and changed-line rule checks passed.
- Old-code falsification through the install entry point failed as expected: the old implementation did not use the managed namespace or evict the completed cache.
- Self-review tightened the byte-limit test so the entry-count limit cannot make that test pass accidentally. Its independent rerun is recorded in the follow-up verification comment.

No UI changes. No release, daily-app restart, legacy-cache migration or profile cleanup is included.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -71,6 +71,7 @@ For maintainers who build, verify, release, recover, or harden o8.
 | [Crash survival](operations/daemon-crash-survival.md) | How worker processes and transcripts recover after app or server failure. |
 | [Project hardening](operations/project-hardening.md) | The multi-repo project contract, retrieval scope, locks, and expected invariants. |
 | [Substrate evaluation gate](operations/substrate-eval-gate.md) | The thresholds and sustainment checks for memory and retrieval quality. |
+| [Dependency cache retention](operations/dependency-cache-retention.md) | Download-cache bounds, active-installer protection, legacy preservation, and maintenance receipts. |
 
 Questions or problems? [Open an issue](https://github.com/hurttlocker/o8/issues) — the bug template asks for the details that make reports actionable.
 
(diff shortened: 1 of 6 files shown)


hurttlocker:
Verification follow-up: the independent byte-limit case passed through the real install entry point with maxBytes=0 while leaving the entry-count allowance at 12. The count, age, concurrency, failure-preservation, identity, symlink and cross-profile cases passed in the preceding 16-test integration run. The branch now incorporates the latest main without conflicts; CI is checking that merged head. No release or legacy-cache deletion is included.

PR #2172, opened 2026-09-11
fix: exclude driver delay from startup interaction timing

hurttlocker:
## Change

Correct the startup-input clock in #1697's interaction harness. A trace of the same released build observed an editable composer 1,520.4 ms after hydration, but the driver read the page at 2,263.7 ms. The old metric counted that 743.3 ms of driver delay as application startup work.

The new page-side probe records editable readiness, then requires the first trusted keyboard input to remain in the same textarea through paint. The metric adds readiness time and input-to-paint time. The excluded driver wait and full hydration-to-paint wall time remain in the receipt. The 2,000 ms ceiling is unchanged.

The probe rejects missing, disabled, readonly, replaced, synthetic or unpainted input and stops after its first observation or deadline. It adds no product code or idle-soak observer. Baseline rows retain the measurement method; an old/new startup comparison reports `incomparable` instead of inventing an improvement. Other metrics retain their previous comparison behavior.

## Verification

Real-browser tests exercise a deliberately delayed driver, a genuine 2,150 ms application delay that still fails the budget, synthetic events, and replacement of the input before paint. The focused harness suite passed 45 tests. Final TypeScript and full-suite results are recorded below before merge.

The previous release observations remain unchanged. Corrected observations are collected separately, not substituted into the old receipts. The budgets remain provisional pending the separate baseline-policy decision. This PR does not claim whole-app performance acceptance or release publication.

### DIFF
--- a/docs/operations/interaction-performance-budgets.md
+++ b/docs/operations/interaction-performance-budgets.md
@@ -102,6 +102,21 @@ design-page digest), and the sample count behind every distribution.
 
 ## The measurement order is part of the contract
 
+Startup input uses `page-readiness-plus-trusted-input-paint-v1`: the page stamps
+when its active textarea is visible and editable after hydration, then verifies
+the first trusted keyboard input remains in that same textarea through paint.
+The metric adds readiness time and input-to-paint time. It excludes the test
+driver's idle/protocol wait between those phases, recording that excluded wait
+and the full hydration-to-paint wall time separately in `observation`. Disabled,
+replaced, untrusted or unpainted input cannot produce a pass. The observer stops
+after the first input or its deadline; it does not run during the idle soak.
+
+Older startup observations ended when the driver read the page. They included
+driver delay and are not comparable to this method. The raw receipts remain
+unchanged; baseline deltas report `incomparable` across a method change. New
+observations must be collected before accepting a replacement startup baseline.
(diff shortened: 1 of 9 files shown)


hurttlocker:
Self-review of `bcd617e3a947aa4be6618384426c5d0dc659bd58` found no blocking defect.

- The probe runs only in the benchmark page, validates a trusted changed input on the same visible editable target, and stops before the soak. App code and numerical budgets are unchanged.
- The receipt retains readiness, trusted input-to-paint, excluded driver/setup wait, and total wall time. The derived startup metric is named and documented rather than presented as an unchanged wall-clock measurement.
- Baselines preserve their measurement method. Cross-method deltas are explicitly incomparable, and matching methods retain normal comparison behavior. Previous receipts are unchanged.
- Final focused real-browser/harness run: 45 passed. Full hermetic suite: 4,315 passed, four skipped. TypeScript, touched-file ESLint, classification and diff checks passed. Rule-check returned no violations and scanned zero application files; this patch changes benchmark scripts, tests and documentation only.

Release observations are being collected separately with the committed instrument. They use exact shipped artifacts, isolated service profiles, and a measured browser. They are not native-window relaunch proof or automatically accepted baselines. Required CI remains the merge gate. No release publication is included.

Issue #2175, opened 2026-09-11
Measure Design Mode screenshot crop through the supported capture path


hurttlocker:
## Remaining measurement

The interaction harness keeps `design_screenshot_crop_ms` explicit and unavailable on the embedded iframe Design Mode path. Both corrected packaged observations (0.1.746 and 0.1.747, scales 50/250/1,000) recorded null with the unsupported-path reason. This is not a zero-duration result or a passing crop check.

The operator approved separating this capability from #1697's accepted v1 harness and baseline deliverable. The existing 500 ms absolute crop ceiling and 40 ms comparison band remain unchanged; unsupported measurements must continue to report unavailable.

## Bounded follow-up

- Resolve the supported capture path for a Design Mode selection in the embedded browser and define the start/end boundaries for crop-ready latency.
- Exercise that path on a local deterministic fixture through real selection and capture, with correct coordinates, scaling and clipping.
- Record a numeric duration only after the crop exists and matches the selection. Keep null plus a reason on unsupported paths.
- Add a deliberately delayed capture that fails the unchanged budget; verify cleanup leaves no fixture browser, process, state or port behind.
- Capture visual before/after proof if the product gains a visible crop attachment.

This issue does not authorize a redesign, a global budget increase, or a claim that native-window and iframe capture are interchangeable.

PR #2177, opened 2026-09-11
test: record accepted Mac interaction baseline

hurttlocker:
## Outcome

Record the operator-approved v1 interaction reference after measuring the exact 0.1.746 and 0.1.747 released artifacts. The manifest is now locked; every numeric ceiling, comparison band, and evaluator branch is unchanged.

The saved reference contains the 45 measured metrics and their provenance. Acceptance remains separate from the raw result: the newer release's 1,083 ms startup at 1,000 rows is accepted as the initial reference while its +153.3 ms comparison failure against the unchanged 150 ms band remains recorded. No automatic acceptance exception was added to the runner.

Screenshot-crop timing stays null and unavailable, tracked in #2175. The existing terminal acceptance remains separate from these release observations. This patch records the approved scope for #1697; it does not claim a native launch benchmark or an eight-hour soak.

## Verification

- Focused harness and real-browser tests: 46 passed.
- Full hermetic suite: 4,324 passed, four skipped.
- TypeScript, touched-file lint, test classification and diff checks passed.
- Changed-line rule check passed with zero application files in scope.
- All 45 saved metric records match the original observation. All numeric budgets, bands and evaluator code match the pre-acceptance manifest.
- The added check confirms that a future +153.3 ms change still fails, and unavailable crop timing still cannot pass.

Required CI receipts will be checked before merge. No version bump, app publication or permanent daily-app replacement is included.

### DIFF
--- a/docs/operations/interaction-performance-budgets.md
+++ b/docs/operations/interaction-performance-budgets.md
@@ -72,9 +72,9 @@ npm run bench:interactions:full -- \
 ```
 
 Each release run writes `tests/bench/results/interactions-baseline-release-<appVersion>.json`,
-so the two coexist and can be diffed. The budgets below stay `provisional`
-(`INTERACTION_BUDGETS.status === 'provisional'`, `lockedBy === null`) until both
-exist and the operator locks them.
+so the two coexist and can be diffed. Exported observations remain `observed`;
+writing an observation does not grant acceptance. The September 11 v1 decision
+below locks the unchanged budget manifest after observing two shipped builds.
 
 Release observations require all three independent provenance links: the
 notarized `.app`, the SHA-256 of its release archive, and a full 40-hex release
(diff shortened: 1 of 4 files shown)


hurttlocker:
Reviewed final head 8466d1b0748e428a77857f3a4fed078a675b5b3c. The PR diff contains only the four interaction-acceptance files; it does not change the planner or other runtime code. The earlier documentation-drift notice refers to separate main-branch work.

The sync from main added only the four already-merged Rust files from #2174. The JavaScript/TypeScript tree is identical to the locally tested acceptance tree (4,324 hermetic tests passed, four skipped; 46 focused harness/real-browser tests passed). TypeScript passed again after the sync. Numeric budgets, comparison bands and the evaluator are byte-for-byte unchanged below the manifest metadata. All 45 saved metric records match the original observation, whose raw `fail` status remains present. No review threads are unresolved.

The native real-profile cache acceptance and daily-app restoration also completed, allowing #1817 to close. That candidate proof is separate from this acceptance-record patch and does not authorize publication. Required CI remains the merge gate for this exact head.

Issue #2142, opened 2026-09-10
Dispatch failure is silent: launch never happens, thread shows a confident mission plan, only a console.warn records it


hurttlocker:
## What happens

A Multitask turn ends with the orchestrator stating, confidently and in detail, that it dispatched three packets:

> Mission: mission-5ff02201-a37. File ownership is disjoint across all three, so the merges should apply cleanly in any order — P1 first is still preferable since it lands `styles.css`.
> Send me a message when you want me to review.

Nothing was dispatched. No worker tiles appear, no lanes exist, no worktrees are created, and the repo never changes. The mission record exists, so `get_mission_status` answers and looks healthy, which makes the claim harder to disbelieve.

o8 detects this correctly and says so in the server log:

```
[orchestrator-session] FALSE-DISPATCH suspected for cortex-orchestrator-…: launchAgentCallCount=0 but text claims dispatch
```

Then it does nothing with that finding. The turn completes normally: `turn.onEvent({ type: 'done', … })`. No error event, no banner, no retry, nothing in the thread. The operator's only signal that the run was fiction is that the repo is still empty some minutes later.

It fired five times in one session of driving this surface.

## The mechanic

`src/lib/lane/orchestrator-session.ts:713` compares two things it already has: the number of launch-agent tool calls in the turn, and whether the assistant's text claims a dispatch. When they disagree it emits `console.warn` and falls through to the normal success path a few lines below (`:718`).

The `narrate-and-exit` check immediately above it (`:708-711`) has the same shape and the same ending: a warn, then success.

So o8 knows the turn was a hallucination and reports success anyway. Everything downstream — the thread UI, the mission surface, any automation watching for lanes — is told the run went fine.

Related in the same trace, and possibly causal: the session was reaped mid-turn under `MAX_LIVE (4)` right before this happened, and the assistant's own text says "I could not hold this turn open until they finish". Worth checking whether reaping a warm orchestrator proc during an active turn is what strands the launch calls.

## Why it matters

A false success is worse than a failure. A failed dispatch is visible and retryable in seconds; a fabricated one costs however long it takes someone to notice the work does not exist, and it teaches operators to distrust the mission surface generally. It is also unrecoverable by automation: anything waiting on lanes waits forever, because the turn said it was done.

## Suggested fix

The detector is already right. Act on what it finds:

[...]

PR #2182, opened 2026-09-11
Name the launch tool in the Multitask and MoA directives (#2153)

### LINKED ISSUE #2153: Multitask mode is a sentence in the user message, not a mechanism: nothing makes the launch tool more likely to be called
Root cause behind #2142. That issue is about the detector reporting success anyway; this is about why there is nothing to detect a success from.

## What Multitask actually does

All of it, in `src/lib/orchestrator/composer-wire.ts:3-7`: prepend one sentence to the operator's message.

```
multitask: '[Mode: Multitask] Decompose this into parallel worker packets and dispatch them
            into isolated worktrees instead of working serially yourself. Review and merge
            through the gate as they finish.'
```

`composeComposerWireMessage` glues that in front of the text and returns. The selected mode never reaches the server as a structured field — grep for `ComposerWireMode` outside `composer-wire.ts` and `composer-mode.ts` and there are no consumers. Nothing keys off it: no tool gating, no `tool_choice`, no system-prompt change, no different turn shape.

So the entire mechanism is: ask, in the weakest position in the prompt, and hope.

## Why it fails the way it does

Three things stack up:

1. **The directive never names the tool.** The model has to infer that "dispatch them into isolated worktrees" means calling `cortex_launch_agent`. Nothing connects the sentence to the capability.
2. **It sits in the user turn**, not the system prompt — the least durable place to put an instruction that must hold for the whole turn.
3. **Nothing verifies it happened.** A model that writes a good decomposition and says "dispatched" has produced a completely reasonable response to that sentence, and the turn is reported as `done` (#2142).

Note the asymmetry with Solo, which works reliably:

```
solo: '[Mode: Solo] Work directly in this session yourself — do NOT dispatch worker agents
       or create missions. Edit, run, and verify with your own tools.'
```

Solo is a **prohibition**, and prohibitions are easy to obey. Multitask is an instruction to use a capability it never names.

## Evidence

From driving this surface repeatedly:

- `POST /api/orchestrator/delegate` works every time — it bypasses the model entirely.
- Explicitly telling the model to call the tool works — it names the tool.
- The normal composer flow does not, across many phrasings, some very decisive ("split it into three packets and run all three in parallel right now").

Server log for a failing turn: proc spawned `full/full`, so the tool was available and permitted; `launchAgentCallCount=0`; `sawToolUseAfterText=false`. No MCP or tool error anywhere. The model wrote its plan and made no tool calls.

## Suggested fix

In rough order of leverage:

- **Name the tool in the directive.** Cheapest change with the largest expected effect.
- **Move the mode directive into the system prompt** for the turn rather than prefixing the user message.
- **Send the mode as a structured field** so the server can act on it, rather than inferring intent from text it wrote itself.
- **Force the call when the mode demands it** — `tool_choice` for the first call of a Multitask turn, or a required-tool turn shape.
- Failing all of that, the turn should verify the mode's contract was met, which is #2142.

## Acceptance

1. A Multitask turn with a decomposable request results in at least one `cortex_launch_agent` call, measured over repeated runs, not once.
2. The selected mode is available server-side as a value, not only as a prefix on the user's text.
3. Solo still refuses to dispatch.
4. A Multitask turn that ends with zero launch calls is a failure, not a success (#2142).

hurttlocker:
Closes #2153.

## What changed

`src/lib/orchestrator/composer-wire.ts` — the Multitask and MoA directives now name `cortex_launch_agent` and state the contract as an instruction rather than advice:

> Dispatch means calling the `cortex_launch_agent` tool — call it once per packet, in parallel, in this turn, BEFORE you write any summary. A turn with no `cortex_launch_agent` call has dispatched nothing, however good the plan is; never claim work was dispatched without the tool results to show for it.

Three supporting details:

- **Bare tool name, on purpose.** Each backend namespaces the cortex MCP tool differently in the model's tool list — Claude sees `mcp__cortex__cortex_launch_agent`, Codex sees `cortex__cortex_launch_agent` / `cortex.cortex_launch_agent` (`proposer-lockout.ts:73`, `orchestrator-session.ts:1088`). The bare registered name (`cortex-mcp-server.ts:307`) is the suffix of all of them, it is what `orchestrator.md` already uses throughout the system prompt, and it is the form the operator's successful experiment used.
- **Solo is untouched** and still prohibits dispatch. It never names the launch tool.
- **`stripKnownComposerWirePreamble` also matches the pre-change directive strings.** It keys on exact directive text, so without this an older client's preamble would stop being recognized and the operator's message would be persisted with the directive glued to the front. `isKnownComposerPreambleTitle` keys only on the `[Mode: …]` marker, which is unchanged; both are covered by tests.

## Evidence this works

From the issue comments, same stage / same mode / same launch policy / same model, the only variable being whether the request named the tool:

- **Named:** three packets dispatched 34 seconds after the message, all three ran to review-ready, the merge gate lit up, and the pipeline finished with three clean merges (`{"merged": 3}`).
- **Not named:** zero launch calls across a full day of attempts and many phrasings, including deliberately decisive ones. `proc full/full`, no MCP or tool errors, `sawToolUseAfterText=false` — the model wrote a plan and called nothing.

Everything downstream of the launch call already works. The directive not naming the mechanism was the whole gap.

## Verification

- `npx tsc --noEmit` — clean, no output.
- `npx vitest run src/lib/orchestrator/composer-wire.test.ts src/components/desktop/thoughts/composer-mode.test.ts src/lib/mobile/orchestrator-thread-history.test.ts` — 3 files, 24 tests passed.
- `npx vitest run tests/solo-send-wire-separation.test.ts` — 1 file, 2 tests passed.
- `npx eslint` on both changed files — clean.

New `src/lib/orchestrator/composer-wire.test.ts` covers: the dispatching modes name the tool, Solo still prohibits and does not name it, every mode's preamble round-trips back to the operator's text, the pre-change directives still strip, and title-marker recognition (current, legacy, bracketed and unbracketed, plus negatives).

## Left out of scope

Deliberately not in this PR, per the issue's own ordering — these are larger architectural changes and each deserves its own review:

- Sending the mode to the server as a structured field instead of a text prefix (issue acceptance item 2).
- Moving the mode directive into the system prompt for the turn.
- `tool_choice` forcing / a required-tool turn shape for the first call of a Multitask turn.
- Failing a Multitask turn that ends with zero launch calls — that is #2142.

This PR is the cheapest change with the largest expected effect, which is what the experiment isolated.

https://claude.ai/code/session_015dVxFgHWcYwiMja5oBxySH

### DIFF
--- a/src/lib/orchestrator/composer-wire.test.ts
+++ b/src/lib/orchestrator/composer-wire.test.ts
@@ -1,0 +1,87 @@
+import { describe, expect, it } from 'vitest';
+
+import {
+  COMPOSER_MODE_DIRECTIVES,
+  composeComposerWireMessage,
+  isKnownComposerPreambleTitle,
+  resolveOrchestratorTranscriptMessage,
+  stripKnownComposerWirePreamble,
+  type ComposerWireMode,
+} from './composer-wire';
+
+const MODES: ComposerWireMode[] = ['solo', 'multitask', 'moa'];
+
+/** Directives as o8 shipped them before #2153 — still on the wire from old clients. */
+const LEGACY_MULTITASK = '[Mode: Multitask] Decompose this into parallel worker packets and dispatch them into isolated worktrees instead of working serially yourself. Review and merge through the gate as they finish.';
+const LEGACY_MOA = '[Mode: Mixture of Agents] After the proposal round, decompose the work into parallel worker packets and dispatch them into isolated worktrees. Review and merge through the gate as they finish.';
+
+describe('composer mode directives', () => {
+  it('names the launch tool in the dispatching modes (#2153)', () => {
+    for (const mode of ['multitask', 'moa'] as const) {
+      expect(COMPOSER_MODE_DIRECTIVES[mode]).toContain('cortex_launch_agent');
+      expect(COMPOSER_MODE_DIRECTIVES[mode]).toMatch(/call it once per packet/i);
+    }
+  });
+
+  it('keeps Solo a prohibition and never points it at the launch tool', () => {
+    expect(COMPOSER_MODE_DIRECTIVES.solo).toContain('do NOT dispatch worker agents');
+    expect(COMPOSER_MODE_DIRECTIVES.solo).not.toContain('cortex_launch_agent');
+  });
+
+  it('keeps every directive prefixed with its [Mode: …] marker', () => {
+    expect(COMPOSER_MODE_DIRECTIVES.solo.startsWith('[Mode: Solo]')).toBe(true);
+    expect(COMPOSER_MODE_DIRECTIVES.multitask.startsWith('[Mode: Multitask]')).toBe(true);
+    expect(COMPOSER_MODE_DIRECTIVES.moa.startsWith('[Mode: Mixture of Agents]')).toBe(true);
+  });
+});
+
+describe('composeComposerWireMessage', () => {
+  it('round-trips the operator text back out of every mode preamble', () => {
+    const operatorText = 'Build a three-page static site: home, about, contact';
+    for (const mode of MODES) {
+      const { displayMessage, wireMessage } = composeComposerWireMessage(operatorText, mode);
+      expect(displayMessage).toBe(operatorText);
+      expect(wireMessage).toBe(`${COMPOSER_MODE_DIRECTIVES[mode]}\n\n${operatorText}`);
+      expect(stripKnownComposerWirePreamble(wireMessage)).toBe(operatorText);
+      expect(resolveOrchestratorTranscriptMessage({ message: wireMessage })).toBe(operatorText);
+    }
+  });
+
+  it('passes slash commands through unshaped', () => {
+    expect(composeComposerWireMessage('/chat Explain this diff', 'multitask')).toEqual({
+      displayMessage: '/chat Explain this diff',
+      wireMessage: '/chat Explain this diff',
+    });
+  });
+});
+
+describe('stripKnownComposerWirePreamble', () => {
+  it('still strips the pre-#2153 directives sent by older clients', () => {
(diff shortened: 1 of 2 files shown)


hurttlocker:
Attempted a pre-merge validation on the demo machine against the live 0.1.747 build, by typing directive text into the composer by hand. One run completed before the machine locked, and it is worth recording because it argues for strengthening this PR.

**Result: naming the tool is not automatically sufficient — position matters.**

Failed (no launch call, turn ended narrating):

> Build a three-page static site in plain HTML and CSS: home, about, contact. Dispatch means calling the cortex_launch_agent tool. Call it once per packet, in parallel, in this turn, before you write any summary.

Succeeded earlier, same stage and model, three packets in 34s:

> Split this into three packets and call `cortex_launch_agent` for each one now: home page, about page, contact page.

The difference is not whether the tool is named — both name it. In the successful one the tool call is the sentence's **primary verb** and the work is its object. In the failed one the request leads with "Build a three-page static site", the model latches onto building, and the tool instruction arrives as a trailing clause it can treat as commentary.

This PR's directive currently opens with "Decompose this into parallel worker packets and dispatch them into isolated worktrees instead of working serially yourself", and the `cortex_launch_agent` sentence lands second. Structurally that is closer to the phrasing that failed than the one that worked.

Worth considering: lead the directive with the tool call as the imperative, and let decomposition be what it is applied to. Something closer to "call `cortex_launch_agent` once per packet" as the first instruction rather than the second.

Caveat, stated plainly: this was typed into the composer, so the text landed in the **user message** after the old directive, whereas this PR puts it in the directive block ahead of the user's ask. Position within the prompt differs, so the test is suggestive, not decisive. I could not run the properly structured version — the demo machine is at the lock screen.

Not blocking the PR on this. The change is a clear improvement over naming nothing at all, and the legacy-directive handling is right. But if we merge and it still does not dispatch reliably, this is the first thing to try.

Issue #1684, opened 2026-08-02
Governed pipeline loses first-diff quality the same two ways every time: over-engineering and missed sub-requirements


hurttlocker:
Benchmarked twice — June 2026 and August 2026 — on the same three real issues from the same base, blind-judged against a pre-registered rubric. Both times the governed pipeline won **zero of three** against the raw model coding alone. Different models each time; the four-trace adversarial review was added between the runs and did not change the outcome.

The result itself is published and accepted: o8's value is the merge gate, not the keystrokes. What makes this worth an issue is that **the same two failure modes produced the loss both times**, two months and two model generations apart. That is structural, not variance.

Governed output scored **8.0 / 8.1 / 8.1** while raw models swung **7.0–9.0**. It reliably clears a bar and never reaches the ceiling.

## Mode 1 — over-engineering

On #1065 (MCP ABI re-exec) the winning diff used **141 added lines**: one focused pre-flight module imported first, relying on ESM's depth-first source-order evaluation. The governed diff used **777 lines**, restructuring the server into bootstrap + implementation and adding a new required bundle to the packaging contract. Both correct. One carries five times the blast radius.

June recorded the identical shape: *"a 666-line refactor where 5 lines would do."*

## Mode 2 — missed sub-requirement

On #1144 (lane reconciliation) the issue named three fixes. The winning diff covered all three in 165 lines. The governed diff covered two, missing the zero-diff relaunch loop.

## Why the pipeline permits both

- `src/lib/orchestrator/self-review.ts` gates on whether anything would **BREAK** the task — typecheck failure, runtime crash, scope creep, security risk. That is a breakage bar, not a quality bar. A correct-but-bloated diff passes it cleanly.
- The adversarial review forces four traces: guard/predicate, scope/partition, sub-requirement coverage, execution-path. Three are correctness traces. **None asks whether the change could be substantially smaller.**
- Nothing enumerates the issue's requirements as a planning step. Coverage is checked after the fact, if at all.

## The two modes are not equally fixable

**Mode 2 is mechanically fixable and worth doing.** Requirement coverage can be a hard gate rather than a trace: extract the issue's requirements into an explicit checklist before implementation, then fail completion if any requirement lacks a citing line. That is deterministic, testable, and independent of model taste.

[...]

PR #2265, opened 2026-09-12
fix: isolate paired coding benchmark collection

### LINKED ISSUE #2252: Paired coding collection changes shared approval policy and launches live missions
## Mechanic

The paired first-diff collector always enters `withTemporaryRequireApproval` and then runs the separate governed end-to-end collection. There is no paired-only entry point. Starting a raw-versus-contract comparison therefore changes the shared app's approval posture and launches additional live missions, even when another operator session is using the app.

This prevents an isolated comparison for #1684 while unrelated workers are active. The relevant path is `scripts/bench/run-coding.ts`.

## Expected

A paired-only collection and judging path uses isolated worker worktrees and immutable receipts, without reading or changing live approval settings or launching governed missions. The existing end-to-end measurement remains separately explicit.

## Acceptance

- [ ] Provide a paired-only preflight, collection, and judging path with no live control-plane dependency or approval mutation.
- [ ] Prove through the command entry point that paired-only mode never calls live app configuration or mission commands.
- [ ] Keep raw and contract arms, task bases, blinding, invalid-arm retention, and the scoring rule unchanged.
- [ ] Use owned dependency clones and run-specific worker identities so archived runs and unrelated workers remain intact.
- [ ] Document the phase boundary and record the change as a dated protocol amendment before collecting new scores.

hurttlocker:
## Mechanic

The coding collector combined the paired first-diff experiment with live end-to-end missions and temporarily changed the shared approval policy. Running that collector alongside ordinary engineering work could change the other session's environment.

## What changed

Add explicit paired-only preflight, collection, judging, and combined commands. They do not access approval settings, probe the app, or launch end-to-end missions. Receipts explicitly mark end-to-end data as not collected.

Paired arms and judges use owned APFS dependency clones and worker names derived from the exact run ID. Existing worktrees are preserved. This hardening applies to paired arms in full collections as well; legacy full and standalone end-to-end phase selection and control-plane behavior remain separate. The paired path requires an owned dependency directory and APFS cloning support.

Integrate the explicit runtime configuration and spawn receipts added by #2262. Update the dated protocol and the first-diff roadmap gap to reflect the published mixed dual-judge results. No new quality result is claimed.

## Verification

- Red: the real CLI rejected the new paired-only phase. A separate regression exposed colliding worker identities for distinct run IDs.
- Green: isolated process-entry fixtures exercise preflight, collection and judging with no live app. They retain the no-control-plane check, owned dependency check, requested configuration and launch receipts, and distinct identities across both formerly colliding run IDs.
- Integrated focused paired/configuration tests: 10 passed. Full local suite: 4,493 passed, 4 skipped.
- TypeScript, touched-file ESLint, rule-check, classification, roadmap and diff checks passed.
- The final ancestry sync retains the exact content tree tested at commit `16f7569d8`; no implementation changed during that sync.
- No live coding benchmark, model-quality measurement or end-to-end mission was run for this change.

Fixes #2252


hurttlocker:
The roadmap row keeps the historical loss reports, the over-engineering and missed-requirement concerns, and direct maintainer ownership. Commit `093a6fa31` qualifies the old headline scores as withdrawn.

The [published benchmark report](https://github.com/hurttlocker/o8/blob/main/docs/user/honest-benchmark-2026-08.md#the-finding-that-matters-most-judges-prefer-their-own-family) explicitly says, “Both results are withdrawn rather than defended.” The later paired contract trial remains mixed and does not clear the decision rule. This preserves the negative evidence without presenting withdrawn comparisons as current proof; #1684 remains open.

Issue #2242, opened 2026-09-12
Successful merge emits a rejected terminal transition after archival


hurttlocker:
## Mechanic

The successful merge transitioned the lane from `merging` to `archived`, then attempted an `archived` to `completed` transition for the `merged` event. The registry refused that terminal-to-terminal transition, and mission status exposed the packet as archived despite a populated `mergedAt` timestamp.

## Expected

A successful merge records one canonical terminal transition without a refused lifecycle event or contradictory status.

## Acceptance

- [ ] The merged event is recorded before archival, or terminal completion is idempotent.
- [ ] Mission status identifies the packet as merged through one canonical field or state.
- [ ] A real-path merge test asserts that no terminal transition is refused.

Issue #2235, opened 2026-09-12
Source web folder picker waits for a native dialog that never opens


hurttlocker:
## Mechanic

On a fresh source run started with `npm run dev`, first-run setup offers **Choose a folder on this Mac**. Clicking it disables the control, opens no visible chooser, and leaves `POST /api/panel/browse-folder` pending for 60 seconds until shutdown.

## Expected

The source web loop either selects a local repository or immediately explains that the native shell is required and gives the supported source-CLI alternative.

## Acceptance

- [ ] The source web loop can add a local folder, or returns a visible native-shell-only message within one second.
- [ ] The button re-enables after failure.
- [ ] The message includes the exact supported command for registering the folder from a source checkout.

Issue #2201, opened 2026-09-11
Tracking: OpenCode 2 and ACP


hurttlocker:
Pillar: Runs on your subscriptions  ·  Status: 100% (5 of 5 shipped)

**Done means:** A non-subscription CLI is a first-class worker and a first-class orchestrator backend.

## Checklist
- [x] #1798 OpenCode readiness reports ready while the resident service is version-skewed — shipped 0.1.717
- [x] #1799 OpenCode service holds packet worktree open and blocks governed close — shipped 0.1.717
- [x] #2117 ACP-backed orchestrator replies render with hard line breaks mid-sentence — shipped 0.1.742
- [x] #2143 OpenCode 2 auth is never detected: o8 reads auth.json, the v2 CLI stores credentials in its database — shipped 0.1.749
- [x] #2194 OpenCode 2 dispatch is refused when 'auth list' succeeds with an empty payload — shipped 0.1.750

## How to help
Open children labeled `claimable` can be claimed by commenting "claiming"; see CONTRIBUTING.md → Claiming work.

hurttlocker:
Every child is closed and released as of 0.1.750, so this arc moved to the roadmap's Shipped list in #2277. Shipped arcs carry no tracking issue.

Issue #2196, opened 2026-09-11
Superseding a mission leaves its packets live, so both missions' work launches together


hurttlocker:
## Mechanic

When a new mission supersedes an earlier one on the same thread, the superseded mission's packets stay live and dispatchable. The server logs the client mission being dropped:

```
[orchestrator-state] POST dropped stale client mission (client=mission-5ef7d474-8b6, server=mission-6457d5bf-f05)
```

but nothing cancels the packets that belonged to the dropped mission. They sit queued behind the dispatch gate, and when the gate finally passes they **all launch at once**.

Observed: an operator re-ran the same ask while the first run was still gated. Three tasks produced **eight packet ids across two missions**, and five redundant lanes launched in a four-minute window — including three separate lanes for the same deliverable with different packet ids. Every one of them had `zeroDiffRuntimeRetries: 0`, `attemptCount: 0`, `launchAttempts: 0`, so this is not a retry path; they are distinct packets from a mission that should no longer exist.

Two of the resulting lanes were then correctly declined by review for duplicate deliverables against work a sibling packet had already merged. The review gate caught the waste, after the compute was spent.

## Why it matters beyond the waste

Re-running an ask is the most natural operator reflex when a dispatch appears stuck — and a dispatch appearing stuck is exactly the condition in #2195. So the two compound: the app looks idle, the operator re-asks, and both missions' packets land together.

It is also the worst possible behaviour on a screen recording: the app shows nothing, then dumps five workers, two of which are rejected for duplicating each other.

## Expected

Superseding a mission cancels its packets. A packet belonging to a superseded mission is never launched.

## Acceptance

1. Creating a mission on a thread that already has a live mission marks the earlier mission's un-launched packets cancelled.
2. A cancelled packet is never dispatched, and its lane is never created.
3. Test drives two missions on one thread and asserts only the surviving mission's packets launch.
4. In-flight packets from the superseded mission are handled explicitly — either allowed to finish or cancelled — and the choice is stated in code, not left implicit.

PR #2277, opened 2026-09-12
docs: record what 0.1.750 shipped on the roadmap

hurttlocker:
## Mechanic

0.1.750 is published and contains the eight first-run fixes (#2235 to #2242), both lifecycle fixes (#2195, #2196), and the OpenCode empty-provider fix (#2194). The tracking checklists are ticked; the roadmap text still described those as pending.

## What changed

- Now: the eight frictions are stated as shipped in 0.1.750; the arc stays open because the download path is unmeasured.
- First ten minutes row: gap now says the fixes shipped and no second run confirms them.
- OpenCode 2 and ACP: row removed from pillar 3 and added to Shipped as 0.1.750, since every child is closed and released. #2201 closes on merge, per the rule that shipped arcs carry no tracking issue.
- Contributor-ready public repo: gap now says every child is shipped and the outcome (an outside pull request through the claiming protocol) is unproven.

## How verified

```
node scripts/roadmap-status.mjs --check
16 tracking issues, 79 of 97 children shipped.
roadmap check: no drift.
```

https://claude.ai/code/session_01E4G8rVjFpi9pdGFW6mALGe

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -16,7 +16,7 @@ Taste is a gate on every row here, not a pillar of its own. A change that reads 
 
 ## Now
 
-One arc first: **First ten minutes.** A new operator goes from download to a first merged packet, and the minutes and the stalls are measured. The first measurement exists: from source on a Mac, clone to merged packet took 6.8 minutes with nine stalls, one of them needing maintainer knowledge to pass. The eight product frictions behind those stalls are the arc's children, and the download path is still unmeasured. [#2211](https://github.com/hurttlocker/o8/issues/2211), [receipt](./docs/user/first-run-receipt-2026-09-12.md)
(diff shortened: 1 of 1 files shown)


hurttlocker (commented):
The 0.1.750 publication has now been independently verified: source and public mirror assets match anonymous downloads, updater and native signatures pass, and the app and installer notarization tickets validate.

Two points before merging this roadmap update:

1. The #2197 row still describes #2195 and #2196 as current gaps. Both are included in 0.1.750 and now checked in the parent. Its remaining unchecked children are #2270 and #2271, so the row should name the snapshot recovery and changed-reviewed-HEAD evidence gaps instead.
2. Moving #2201 to Shipped also asserts its **Done means** outcome: the runtime works as both a worker and an orchestrator backend. Please link the existing real-entry-point evidence for both roles in the closeout. All five children being released proves the shipped-child count; it does not by itself establish that broader outcome. If that evidence is still absent, retain the open row and name the remaining verification.

The #1684 row correctly retains the negative fixed-trial result and unmet decision rule.

PR #2333, opened 2026-09-13
feat: composer selector is the default; work mode and effort are independent (#2315)

### LINKED ISSUE #2315: Separate work mode from reasoning effort in the composer; the top slider notch writes three settings
Part of #2314.

## Mechanic

The reasoning-effort slider's final notch is a work mode disguised as an effort level. Picking it writes three settings in one click: it enables the parallel deep pass, disables the two-plan comparison mode, and sets effort to xhigh (`src/components/desktop/thoughts/ModelThinkingChip.tsx`, `handleEffortPick`). The composer then resolves the deep pass over whatever work mode was chosen (`resolveComposerExecutionMode` in `src/components/desktop/thoughts/composer-mode.ts`), and a turn composed as Solo with the deep pass armed is sent to the model with the Multitask directive.

Two controls on the same composer then disagree. The effort chip's mode label reads from the deep-pass flag while the mode picker still shows Solo, Multitask, or the comparison mode.

## Expected

Work mode and reasoning effort are independent axes. The deep pass is a work mode next to Solo, Multitask, and the comparison mode, not a slider stop. Picking any effort level never changes the work mode. Picking any work mode never changes effort. One resolver produces the at-rest label, the mode indicator, and the wire directive, so they cannot disagree.

## Acceptance

A real-path test drives the composer's turn composition entry point with every (mode, effort) pair and asserts:
- the wire directive matches the chosen mode for every effort level, including the top one;
- the effort value survives every mode change;
- the rendered mode label and the wire directive come from the same resolver output.
A second test renders the chip and the mode picker together and asserts they never show different mode names for the same state.

This reverses the July 2026 choice that put the deep pass at the top of the slider. The reason for reversing it is above: the coupled notch makes the next turn unpredictable from the composer.

hurttlocker:
Closes #2315

## What changed

- The composer selector now defaults on. Storing 0 keeps the classic footer available.
- Work mode and reasoning effort now resolve independently on both footer variants.
- Fusion is a fourth work mode, and the legacy swarm preference migrates once into the per-tab mode setting.
- The classic effort control now uses the shared supported-effort resolver. Its coupled top notch is gone.
- The default footer preserves repo targeting, session rules, context meter, leading controls, voice mode, and compact-width measurement.

## Verification

Red before implementation:

```text
Test Files  1 failed | 1 passed (2)
Tests  2 failed | 32 passed (34)
```

```text
Test Files  1 failed (1)
Tests  no tests
Error Failed to resolve import "./composer-mode-storage"
```

Green:

- `composer-selector/state.test.ts`: every mode and effort pair reaches the route resolver; effort survives mode changes.
- `composer-mode.test.ts`: labels, directives, and execution modes share one resolver.
- `composer-mode-storage.test.ts`: legacy migration persists once and stored mode wins.
- `ComposerSelectorFooter.test.tsx`: default/opt-out surfaces, classic top effort, Fusion independence, and preserved default-footer controls.
- `fresh-turn-options.test.ts`: migrated Fusion mode reaches the real websocket payload.
- `npx tsc --noEmit`
- `npx vitest run src/components/desktop/thoughts`: 41 files, 227 tests passed.
- `npm test`: 723 passed, 3 skipped files; 4,538 passed, 4 skipped tests.

https://claude.ai/code/session_01E4G8rVjFpi9pdGFW6mALGe


hurttlocker:
Holding merge for one review finding: a model or backend pick while the composer is in the comparison or deep-pass mode must reset the mode to Solo through the single mode setter, otherwise the stale comparison flag still forces the turn's backend (regression of the routing fix in 5e767258d). Fix and a real-path test are in progress on this branch. Second, smaller item: the classic chip resolves the effective lead model differently from the selector footer, so it can hide the top Codex level when no model is pinned.

PR #2335, opened 2026-09-13
Show effective execution receipts on orchestrator turns

### LINKED ISSUE #2316: Show what actually ran on every sent turn: lead model, effort, work mode, worker runtime
Part of #2314.

## Mechanic

A sent orchestrator turn records elapsed time, tool count, files, tokens, and cost (`src/components/desktop/thoughts/chat-panel/TurnSummaryCard.tsx`), but not the lead model, reasoning effort, work mode, or worker runtime it ran with. The only place the effective model is visible is the chip at the moment of sending. Once the turn is in the transcript there is no way to tell what ran, and when the composer overrides a mode (see the coupling issue) the operator never learns it.

## Expected

Every sent turn carries a compact receipt: lead model, effort, work mode, and worker runtime when workers were dispatched. The receipt shows the effective values, meaning what the backend actually applied, not what the chip displayed. When the effective mode differs from the picked mode, the receipt says so.

## Acceptance

- Turn metadata persists the four values with the turn, and a reload renders them from persisted state, not from current composer state.
- A real-path test sends a turn through the orchestrator turn path with a known model, effort, and mode, then reads the transcript and asserts the receipt values.
- A second case sends Solo with the deep pass armed (or whatever override remains after the coupling fix) and asserts the receipt shows the effective mode.
- Worker turns dispatched from that turn show the runtime and model each worker ran, so the receipt is not only about the lead.

hurttlocker:
## Mechanic

Sent orchestrator turns did not retain one durable receipt for the effective lead model, effort, work mode, and launched worker runtime/model. Reloaded desktop and mobile history therefore could not show what actually ran.

## What changed

- Persist the effective turn receipt at send time and preserve it through chat-history normalization, archive, and mobile projection.
- Attach launched workers to the originating thread and turn after the real packet launch succeeds.
- Merge early worker attachments safely when assistant text arrives later.
- Render the persisted lead and worker execution details in the turn summary.
- Exercise the worker attachment through a real WebSocket turn, mission dispatch, managed worktree, and fake worker launch.
- Give the real-path fixture a bare origin with `main` and an affirmative `/api/setup/status` response, matching the production launch preconditions.

## Verification

Red receipt before the fixture correction:

`3/4 passed`

`Dispatch backend is still warming after 20003ms (http_404); retry the launch once the o8 server is ready.`

The failure took the `runtime_launch_catch` branch before the turn worker append.

Green receipts:

- `npm run test:integration -- tests/orchestrator-model-attribution-real-path.test.ts -t 'worker receipt'`: 4/4 passed.
- `npx tsc --noEmit`: passed.
- `npm test`: 723 files passed, 3 skipped; 4,546 tests passed, 4 skipped.
- `npm run rule-check -- --base=3d5ca84bd`: zero violations.
- Changed-file ESLint: zero warnings.

No visual proof was captured because the real-path verification owns processes and persisted state; o8 self-UI review remains on the operator surface.

Fixes #2316


hurttlocker:
Review pass (independent reviewer) found one blocker and two nits.

**Blocker, fixed in bee3f4541:** the chat-history PATCH handler did its full read-modify-write outside the new history lock, so a star or rename racing a turn-receipt write could put the pre-race snapshot back on disk. PATCH now holds the lock like POST does. New integration case in `tests/chat-history-merge.integration.test.ts` drives the real route while another process holds the lock; it fails without the fix (route resolves before the worker row lands) and passes with it.

**Nits, left as follow-ups:** the lock waits synchronously up to 5s under contention, which would stall the WS process if writes ever got slow; `readOpenclawAgentPrimaryModel` has no dedicated unit test.

**Known residual (unchanged):** the ws-server rebound handler and the supervisor auto-queue drain ignore `turn_receipt`, so turns replayed through those paths carry no receipt.

Also since the last CI run: merged current main (mission staging landed on the same handler), regenerated the test-classification manifest that the new integration file was missing from, and re-ran the four-case real-path suite on the merged tree: 4/4.

PR #2337, opened 2026-09-13
feat: centralize composer selector state and labels

### LINKED ISSUE #2319: One resolver for selector state and one label table per setting
Part of #2314.

## Mechanic

Selector state lives in four places: per-thread stored state, per-tab persistence through `onModePersist`, direct localStorage keys (`o8:orchestrator:model-<repo>`, `o8:orchestrator:thinking-preference-*`, `o8:orchestrator:swarm-*`, `o8:orchestration-mode-<repo>`), and the operator-defaults API for worker runtime, worker model, and worker start mode. Which one wins on a fresh session versus a restored thread is decided control by control.

The same setting also renders under different names: the fifth effort level is "Extra" in the chip and "xhigh" in `InputButtons.tsx`; the comparison mode is "Mixture of Agents" in the picker and "MoA" on the chip; the worker start mode is "Run now" and "Run".

## Expected

One resolver decides selector state precedence (in-session override, then thread, then operator default) and every control reads and writes through it. One label table per setting, and every surface that renders the setting uses the table. Short and long forms are allowed, but they come from the same row.

## Acceptance

- A test walks the label table and asserts every rendered string in the composer, the chips, and Settings matches a row.
- A real-path test restores a thread with stored selector state, then changes a control in-session, then opens a fresh thread, and asserts each of the three reads the expected value.
- No control writes to localStorage or the operator-defaults API directly; grep-level check in the test.

hurttlocker:
## Mechanic
- resolves composer mode, effort, and worker defaults through the shared selector controller
- derives effort, mode, and worker-start labels from shared long and short label tables
- routes selector persistence through shared storage and operator-default client seams
- restores persisted per-repo models through a side-effect-free callback, separate from deliberate lead picks
- prevents model restoration from resetting Compare plans or Fusion to Solo
- removes the unused selector wrapper so tests exercise the real ComposerArea path
- keeps effective-turn receipts on the same shared effort and mode vocabulary

## Review follow-up
- short mode chip text remains `MoA`; its aria-label and title use `Compare plans`, and the menu renders `Compare plans`
- RED before the callback split: the named real-path restore test failed with `expected solo to be fusion`
- GREEN after the split: `ComposerSelectorModelRestore.test.ts` passed 1/1 and preserved Fusion plus tab storage across a repo change

## Verification
- `npx tsc --noEmit`
- `npm test` — 728 files passed, 3 skipped; 4,576 tests passed, 4 skipped
- focused selector and ownership run — 3 files, 29 tests passed
- deliberate lead-pick send-seam run — 6 tests passed
- `npm run test:classify` — 730 hermetic, 429 resource-owning; the new restore test is hermetic
- `npm run rule-check -- --base=HEAD`
- `npm run rule-check -- --base=origin/main`
- touched-file ESLint — zero errors; two existing warnings in ComposerArea
- independent review line-ceiling finding addressed with no net line growth in the legacy panel

`npm run test:integration` was not run because the added test is hermetic and owns no external resources.

Fixes #2319


hurttlocker:
Held for one fix round after my read and an independent reviewer pass.

**Must fix:** the hook's stored-model restore effect fires on any `modelId` change and goes through the user-pick path. Two consequences: a repo switch or an operator-defaults refresh can reset a thread from Compare plans or Fusion back to Solo and overwrite its stored mode, and accepting a backend handoff (whose apply no longer writes storage) gets reverted to the stale stored model. Restore must be side-effect free and scoped to mount, repo change and thread change; every accepted pick path must leave storage and session agreeing.

**Test gaps:** the storage-ownership test scans three files and four key prefixes and misses the composer-mode key entirely; it becomes an allowlist walk over the whole thoughts tree. Real-path precedence coverage exists only for effort; mode and worker settings get the same fresh-vs-restored test through the real composer.

**Also:** the mode chip's short form is now `MoA` where v1 showed `Compare plans`; keeping it, with the long form asserted on the label and menu. The non-View footer wrapper has no product callers left.

PR #2365, opened 2026-09-14
fix: retire confirmed missing worktree entries

### LINKED ISSUE #2352: Worktree cleanup refuses durable retirement forever when the directory is already gone
## Mechanic

`src/lib/worktree/manager.ts` worktree cleanup begins a durable retirement for every metadata entry it wants to remove. When the entry's directory is already gone, `lstat` returns `ENOENT`, the cleanup logs `[worktree-cleanup] REFUSED durable retirement begin for <path>: ENOENT` and leaves the entry in place. The next cleanup pass repeats the same refusal for the same entry.

Observed on 0.1.753 with 17 `status: "ready"` entries whose directories no longer exist: every cleanup pass logs two lines per entry (`LIVE-PROCESS GUARD OVERRIDDEN` then `REFUSED durable retirement begin`), forever. The metadata never converges, and the log noise buries the real failure of the same run.

## Expected

An entry whose directory is absent is treated as already retired: the metadata entry is removed (or marked retired) with a reason, once. Directories can disappear outside o8 (an operator cleanup, a restore from backup), so the cleanup must converge instead of refusing.

## Acceptance

- [ ] A real-path test seeds a metadata entry, deletes its directory, runs cleanup twice, and asserts the entry is gone after the first pass and the second pass logs nothing for it.
- [ ] `npx tsc --noEmit`, the touched suites, and `npm test` pass.

hurttlocker:
## Mechanic

Cleanup of a registered packet workspace repeatedly refused after its tracked directory disappeared. Snapshot capture tried to read the missing checkout before exact retirement could confirm that its owned parent remained valid, so the metadata entry survived every pass.

## What changed

Ordinary cleanup can proceed only after a read-only metadata check verifies the durable parent and canonical child authority, and the child probe reports ENOENT. Other filesystem errors and ownership uncertainty preserve the entry. Existing snapshot retirement and replay keep their state-machine checks; PR, merge, and discard capture remain strict when verified snapshot evidence is absent.

The lane audit separates an observed missing directory from completed retirement. A completion event follows metadata removal, so a later cleanup failure cannot leave a false completion claim. A second cleanup pass is quiet, and unrelated worktree state stays intact. Without a verified snapshot, cleanup retains the Git branch so deleting a directory cannot also erase the only named reference to unmerged work.

## Verification

- Regression on unchanged main: 1 failed, 4 passed. The real `WorktreeManager.cleanup` call returned false for the confirmed-missing registered workspace.
- Added authority and late-failure controls: 4 failed before correction, with legitimate existing-snapshot replay passing.
- A real-Git preservation probe failed before the branch-retention guard: metadata removal succeeded but the unique commit lost its named reference. The final test checks the branch SHA and committed content through two cleanup passes.
- Corrected focused suites: 34 passed across the new 17-case cleanup suite, managed-creation crash recovery, creation-retirement refusal, operator lifecycle retirement, and exact-directory retirement.
- `npx tsc --noEmit`, touched-file ESLint, changed-line rule check, and test classification passed.
- `npm test`: 4,599 passed, 4 skipped (730 files passed, 3 skipped).

All filesystem operations used disposable fixtures. No visual proof applies to this backend change. No release or installed-app verification is included.

Fixes #2352


hurttlocker:
A final preservation check found a blocker in the current candidate: after a registered Git worktree directory is removed, its branch can still be the only named reference to an unmerged commit. The newly reachable cleanup path removes metadata and then force-deletes that branch without having run committed-work preservation.

A disposable real-Git probe reproduced this: cleanup returned success, but only the main branch remained in `git for-each-ref`; the unique commit had lost its named reference. Holding this PR as draft until metadata cleanup retains the branch or creates verified recovery evidence, with regression coverage through a second cleanup pass.

hurttlocker:
The branch-preservation blocker is repaired. Cleanup now retains the branch when a missing checkout has no verified snapshot, preserving surviving committed work while removing stale metadata. The new real-Git regression checks the branch SHA and committed file through two passes.

Final local verification: 34 focused tests passed, including the 17-case regression suite; typecheck, touched lint, rule check, classification, and the 4,599-test hermetic gate passed. Returning this PR to review and waiting for CI on the updated commit before merge.

Issue #2260, opened 2026-09-12
Three hermetic suites fail intermittently under full-suite load and pass in isolation


hurttlocker:
## Mechanic

Across four full `npm test` runs on 2026-09-12 (PRs #2256, #2258 and their reruns), the same files failed under full-suite load and passed when run alone on the same head:

- `tests/stop-client-state-race.test.ts` (failed in two separate runs)
- `src/lib/problem-dossiers-activity` suite (one run)
- `coding-cli-mission-policy.test.ts` (timeout, one run)

Each rerun of the failed files in isolation passed 100 percent. The failures are timing-sensitive, not logic regressions, and they cost every PR a manual rerun to prove it.

## Expected

The hermetic suite is deterministic under the bounded-worker configuration in `vitest.config.ts`. A test that depends on wall-clock timing either awaits the real signal or moves to the resource-owning suite via `npm run test:classify`.

## Acceptance

- [ ] Each of the three files passes ten consecutive full-suite runs on CI hardware, or is reclassified as resource-owning with the reason recorded in `tests/test-classification.json`.
- [ ] Any fixed timing dependency is replaced by awaiting the event it was waiting for, not by a longer timeout.

hurttlocker:
Another pre-existing set, seen while verifying #2284 against origin/main at 0a2989a54: five cases in `tests/reset-stop-registry-truth.test.ts` fail with expected 200, received 400 from the dispatch fixture, identically on main and on the fix branch. Not caused by the #2224 change; adding it here so the suite-health list is complete.

hurttlocker:
Rechecked `tests/reset-stop-registry-truth.test.ts` while validating the retry-archival repair for #2313. Six cases fail identically on unchanged main tree `6ac2a5993` and on the repair candidate: the five dispatch-fixture 400 responses already recorded above, plus the outgoing-mission close case returning 409 `session_archive_unconfirmed` because recovery metadata cannot be detached. Both runs finish 28/34. The retry salvage, handoff, and stale-generation controls in that file pass.

These are baseline failures, not a green integration run. The new reset-route regression and its archive-refusal, same-key replay, and archival-race checks pass separately. No production gate was disabled to accommodate this baseline.

hurttlocker:
[...]

PR #2394, opened 2026-09-16
feat: route rejection and steer reasons to the next worker and the Brain

### LINKED ISSUE #2219: Rejection and steer reasons are recorded but never reach the Brain or the next worker
## Mechanic

When the operator rejects a packet or steers it, the free-text reason is stored on the approval event for the audit trail. None of the Engineering Brain retrievers read approval events (they index directives, session outcomes, pull requests, issues, and docs), and the session-outcomes column that was designed to flag operator rework is declared in the schema but written nowhere in the codebase.

So the one signal that says "the operator disagreed, and here is why" is captured and then unreachable: a later question to the Brain, or a later worker on the same file, never sees it. This is the reachability pattern the repo's own doctrine warns about: the path exists on paper and nobody can get to it.

## Done means

- A reject or steer reason is written to a row the Brain retrieves from (either a retriever over approval events, or the outcome row's rework flag plus the reason text in its summary).
- A real-path test drives a rejection through the actual route handler, then asks the Brain about that file and asserts the reason appears in the cited sources.
- A dispatched worker's context block for the same repo includes recent operator corrections, not only outcomes.

## Non-goals

No new UI. This is the write path and the retrieval path.

hurttlocker:
Closes #2219.

## Mechanic
The approvals route discarded the `reason` sent with a reject, so the operator's rejection reason was never persisted. Steer messages were recorded on the lane but nothing read them back. Neither reached the next worker on the packet or the Brain.

## Change
- The reject path stores the trimmed reason (capped at 4000 chars) on the approval's resolution note.
- One reader (`src/lib/cortex/operator-corrections.ts`) turns rejected approvals with a note plus `steered_packet` lane events into corrections, newest first, bounded by a scan cap and a limit, scoped by resolved repo path.
- `buildPacketPrompt` adds an "Operator corrections for this packet" block for the next worker dispatched or rerun on the same packet (up to 5).
- `buildContextBlock` adds a "Recent Operator Corrections" section (up to 3) for other packets on the same repo, excluding the packet's own corrections, dropped before directives when the budget is tight.
- A new Brain retriever surfaces corrections under citation kind `correction` with authority 0.95 (below directives, above outcomes). The two citation surfaces gained the one label entry the type required.

## Evidence
- Real-path test `tests/operator-corrections-real-path.test.ts`: drives the real reject route handler with a constructed request, then the real `buildPacketPrompt` for a rerun, then the real `retrieveAll` and `unionMerge`. Red on main (3 of 4 failed: reason never stored, prompt missing the block, retriever empty), green after (5 of 5).
- `npm run test:integration -- tests/operator-corrections-real-path.test.ts`: 1 file, 5 tests passed. Manifest regenerated.
- `npx tsc --noEmit` clean.
- `npm test`: 4613 passed, 2 failed, 4 skipped. The two failures (`stop-client-state-race`, `coding-cli-mission-policy` timeout under load) fail on the base commit as well and are the subject of #2260.

## Residual
- The steer test writes the same `steered_packet` event `steerPacket` records rather than calling the steer route.
- The Brain test proves retrieval and merge, not that the answer model cites the reason in prose.


hurttlocker:
Reviewer pass: no blockers. Two residuals recorded for follow-up, not fixed here: the citation display-title switch has no explicit correction case (resolves through the default branch today), and steered_packet events are written without an operator-only gate, so a steer raised by a non-operator principal would surface as a correction. Steer authorship is unchanged by this PR.

hurttlocker:
Steer authorship residual filed as #2395.

PR #2396, opened 2026-09-16
fix: rate merge-mechanics approvals from the changed paths

### LINKED ISSUE #2381: Base-moved packets present an impossible fast-forward as an approval decision, and docs-only diffs rate High risk
Hit while dogfooding the mobile Review surface.

## What the operator sees

A card sitting in mobile approvals for six days, labelled **High risk**, titled
"Fast-forward blocked: ... (base moved)", with Approve / Request changes /
Deny & stop the agent. Nothing on the card says what the work is, so the
operator cannot tell what they are accepting.

The underlying change is one commit adding one new file, 60 lines of
documentation, no code touched. It merges into `main` with no conflict.

## Two defects

**1. The merge can never succeed, and the gate asks anyway.** The packet
branched from a base `main` is now 132 commits ahead of. The final step is a
fast-forward, which by definition requires `main` to still be at the branch
point. Any packet that sits long enough becomes permanently unmergeable, and
the failure is then presented to the operator as a decision. Approving it
again just retries the same impossible fast-forward.

Either the approve path should rebase or merge rather than fast-forward, or a
base-moved packet should surface as "needs a rebase" with a button that does
it, not as an approval gate.

**2. A docs-only change is rated High risk.** One added file under `docs/`,
zero code, rated the same as a change that touches the runtime. That trains the
operator to ignore the risk label, which is the one signal that has to stay
trustworthy.

## Acceptance

A packet whose base has moved offers a rebase path instead of a doomed
fast-forward, and a diff touching only `docs/` does not come back as high risk.

hurttlocker:
Closes #2381 (second defect; the first was corrected in the issue thread and its follow-up landed in #2384).

## Mechanic
`createFastForwardFailureApproval` and `createRebaseConflictApproval` set `risk: 'high'` from the failure category alone. The diff files were already parsed and attached to the approval but never consulted for risk. A base-moved packet whose whole diff was one markdown file under `docs/` reached mobile as "High risk" beside a gate result of `passed: true, violations: []`.

## Change
- New helper `riskForChangedPaths(files, fallbackRisk)`: every changed path under `docs/` or ending in `.md` rates `low`; anything else, or an empty file list, keeps the fallback.
- Both merge-mechanics builders set `riskFromChangedPaths: true`; `createLaneActionApproval` applies the helper where the diff is parsed. Titles and summaries still name the mechanics reason.

## Evidence
- `tests/worktree-side-merge-real-git.test.ts`: a git hook moves `origin/main` each time a rebased candidate arrives, so the real merge escalates after its retries. Docs-only case: red on main (`expected 'high' to be 'low'`), green after. A `.ts` case keeps `high`.
- `src/lib/lane/changed-path-risk.test.ts`: docs/ only, .md only, mixed, and empty list.
- `npx tsc --noEmit` clean. `npm run test:integration` on the file passes.
- `npm test`: 4600 passed, 2 failed; the failures are the #2260 flaky suites and fail identically without this change.

## Residual
A dirty-working-tree failure with a docs-only diff also rates low now; the failure category still names it in the title.


hurttlocker:
Reviewer pass: no blockers. Residual noted: a rename is classified by its destination path only, so a rename of a code file into a docs path with no content change would rate low. Not the base-moved shape this fixes.

PR #2398, opened 2026-09-16
fix: keep server-persisted packets when a stale client mission snapshot is posted

### LINKED ISSUE #2351: Delegated packet is refused as not found in durable state when a concurrent lane fails during the same turn
## Mechanic

`resolveLaunchWorkMode` (`src/lib/runtime/launch-work-mode.ts`) refuses a launch when the packet is not found in the persisted control-plane state or the mission registry: `Dispatch refused: packet <id> was not found in durable state, so its work mode is unknown.` That is the right fail-closed default.

Observed on 0.1.753 inside one orchestrator turn: the delegate route (`/api/orchestrator/delegate`) created the packet, opened its lane, and the launch was refused with that message. The lane went `launching -> idle`. Two other launches in the same turn had just failed at worktree provisioning. The delegate route appends the packet by mutating state inside `withLockedState`; a concurrent failure path in the same turn appears to have written a state that no longer contained the new packet before the launch read it.

Only one refusal was seen; the failing conditions were a burst of six delegated launches with two provisioning failures interleaved.

## Expected

A packet the delegate route just persisted is found by the launch that immediately follows it, regardless of other lanes failing concurrently. If the packet can legitimately vanish between persist and launch, the refusal should say which writer removed it.

## Acceptance

- [ ] A real-path test drives the delegate route for two packets while a third lane fails provisioning concurrently, and asserts both launches resolve their work mode (found) rather than refusing.
- [ ] The refusal, when it does fire, includes the last control-plane writer or mission id consulted.
- [ ] `npx tsc --noEmit`, the touched suites, and `npm test` pass.

hurttlocker:
Closes #2351.

## Mechanic
The dashboard's lane-reconcile effect POSTs its cached mission to the orchestrator state route whenever lane state changes, including when a sibling lane fails. The server merge kept only the packets that snapshot listed. A packet the delegate route persisted after the client cached its mission was erased from durable state, so the launch was refused as "not found in durable state" and `close_packet_unmerged` reported it missing. Seen twice on 0.1.753 and 0.1.754.

## Change
- Server: a packet missing from a client snapshot is never deleted. Only an explicit `removedPacketIds` entry removes one, and a stopped or reset packet stays protected either way. Lock scope unchanged.
- Client: the mission-change handler records the packet ids it actually removed and sends them with the debounced POST. The packet row's Delete still works.

## Evidence
- `tests/delegate-packet-stale-client-write.test.ts`: real delegate route creates the packet, a stale snapshot goes through the real state route mid-turn, then the real launch resolver and the real discard-packet route. Red on main with the production message (`Dispatch refused: packet pkt-delegate-… was not found in durable state`, 2 failed), green after (2 passed; close returns 200 with the packet archived).
- `npx tsc --noEmit` clean. `npm test`: 4613 passed, 4 skipped, 0 failed on the run before commit; `stop-client-state-race` failed on later reruns and on the base commit under load (#2260).
- Test classified hermetic; manifest unchanged.

## Residual
- The issue also asks the refusal to name the last writer. Not added here.
- The cross-process control-plane lock gives up after its budget and proceeds without the lock by design, logging loudly. A slow holder in the other process could still lose an update; filed separately.


hurttlocker:
Lock fail-open residual filed as #2399.

hurttlocker:
Reviewer pass: no blockers. Every client deletion path routes through the mission-change handler that records removals; the lane-reconcile effect maps packets one to one and never deletes, and server-side consumers do not post full snapshots. Nit noted: store.ts was already past the 800-line ceiling before this change.

PR #2400, opened 2026-09-16
test: remove wall-clock races from four hermetic suites

### LINKED ISSUE #2260: Three hermetic suites fail intermittently under full-suite load and pass in isolation

hurttlocker:
Closes #2260.

## Mechanic
Four hermetic suites passed in isolation and failed under full-suite load with bounded workers. Each had a wall-clock assumption:

- `stop-client-state-race`: two same-tick dynamic imports of a mocked module share the runner's importer call stack, so the second stop skipped the mock factory and evaluated the real mission-service graph, missing the kill wait.
- `stop-fast-ack`: the first `stopPacket` call loaded the mocked service's original graph inside the 5s ack deadline.
- `coding-cli-mission-policy`: the MCP handler graph was imported inside the test body, spending most of the 30s timeout on transform.
- `problem-dossiers-activity`: fixed 850ms and 20ms sleeps raced the 750ms receipt poll and the inbox fetch.

## Change
Test files only. The second stop starts once the first is inside kill confirmation and the tests await kill-entry signals instead of polling; the service graph is pre-loaded at collection time; the handler is imported statically; the activity tests wait for the rendered result instead of sleeping. No production source touched, no timeouts widened, no retries, no reclassification.

## Evidence
Worker report follows in a comment with the bounded-worker reproduction tail and three consecutive passing runs.


hurttlocker:
Reviewer pass: no blockers. Independent run in the worktree: the four files together, 15 of 15 passed; full npm test under the bounded-worker config, 731 files and 4611 tests passed with exit 0; stop-client-state-race alone, 8 of 8. No production source touched, no reclassification. Nit recorded: renderUntil has no iteration cap, so a real regression surfaces as a test timeout rather than an assertion failure.

PR #2401, opened 2026-09-16
fix: limit durable review coverage to the reviewed commit

### LINKED ISSUE #2254: Durable review coverage can count edits outside the reviewed commit
## Mechanic

`listChangedPathsForCoverage` in `src/lib/lane/durable-review-approval.ts` combines the committed packet range with `git diff --name-only HEAD`. The durable review is pinned to the current commit, but a citation can therefore be satisfied by an uncommitted edit absent from that commit.

There is a related publication path to verify: the non-snapshot branches in `src/lib/lane/commands.ts` and `src/lib/lane/worktree-side-merge.ts` can call `commitDirtyWorktree` after the durable-review or reviewed-head check, then publish the resulting commit. The snapshot-specific fingerprint guards do not cover every such branch.

This is a source-traced finding, not yet a reproduced publication. It predates #2248 and is separate from that PR's path-equivalence fix. It matters to #1684 because a review-approved HEAD is used as the final-output measurement artifact.

## Expected

Requirement evidence and any automatic publication authorization must refer to the same immutable bytes. A commit-only review cannot gain coverage from later dirty edits. If a dirty snapshot is intentionally reviewable, its fingerprint must remain authoritative through commit and publication.

## Acceptance

- [ ] Add a persisted-review and real-Git regression where an evidence path exists only in dirty edits.
- [ ] Trace both PR creation and merge from the actual command entry point; prove whether later dirty bytes can be published under an earlier review.
- [ ] Reject stale evidence or require a verified snapshot/new review before publishing changed bytes.
- [ ] Preserve intentional operator-approved snapshot behavior and test a valid control.

hurttlocker:
Closes #2254.

## Mechanic
A durable approved review is pinned to a HEAD SHA, but coverage unioned the reviewed commit range with the live working tree, and the merge and PR paths auto-commit dirty edits after the review check. Edits made after the review could satisfy contract evidence and ride the approved publication.

## Change
- Coverage reads the reviewed commit range only.
- Before the coverage gate, the assessor lists uncommitted paths (tracked and untracked, ignores respected). Any present withholds authorization with a reason naming up to ten of them; agent merges then fall to the existing operator approval card instead of committing unreviewed bytes. If git cannot answer, authorization is withheld (fail closed).
- Nothing is dropped or auto-committed by this change.

## Evidence
Worker report follows in a comment: red tail on main, green tail, integration run of `tests/durable-review-uncommitted-edits-real-path.test.ts`, typecheck, and full-suite counts. Manifest updated for the new resource-owning test.

## Residual
A packet worktree holding untracked, non-ignored files (a stray report, for example) now needs an operator card before an agent merge. That is the intended fail-closed behavior; the reason names the paths.


hurttlocker:
Worker evidence (Opus 5 seat, commit 938ea3e34):

- New real-path test drives `dispatch({verb:'merge', actor:'orchestrator'})` against real git with a stored review. Red on unmodified source: `AssertionError: expected true to be false` at the coverage assertion, 1 of 2 failed. Green with the fix: 2 of 2. The refused merge names `notes.txt` and `extra.txt`, main and origin stay at base, the worktree HEAD stays at the reviewed SHA, and a clean worktree still merges.
- Existing suites with the fix: durable-review-approval 11/11, contract-coverage (includes the #2248 case) 12/12, review-stall-reconcile 13/13, automatic-release 14/14, quota-fallback 8/8, steered-second-pass 2/2, thin-workspace 3/3.
- `npx tsc --noEmit` clean. `npm test`: 731 files passed, 3 skipped; 4611 tests passed, 4 skipped; exit 0. Manifest regenerated.

Residual: three integration suites (merge-checkout-coupling, require-approval-merge, worktree-side-merge-real-git) timed out locally under a load average above 300 and are being re-run now that load is down; CI's integration job is the arbiter. Three other suites fail identically on the unmodified source (reset-stop-registry-truth, steered-review-settlement, storage-admission-rerun-release) and are not caused by this change.

Behavior change to note: an agent merge from a dirty worktree now needs a commit and a fresh review, or the operator card. Operator-actor merges are unchanged.

hurttlocker:
Reviewer pass: no blockers. Merge order confirmed: the durable review is assessed before the merge gate and before any dirty-tree commit on both the merge and create-PR paths, so unreviewed edits are refused by name, not committed. Residual noted, pre-existing: the uncommitted-path check and the merge are not atomic, so a concurrent writer to the worktree between them could still slip in; this change narrows that window.

PR #2402, opened 2026-09-16
test: let the repo switch commit before releasing the gated refresh

### LINKED ISSUE #2397: fresh-turn-options unit test flakes: a cancelled default refresh still emits an orchestrator send under CI load
## Mechanic

`src/components/desktop/thoughts/use-orchestrator-stream/fresh-turn-options.test.ts`, case "cancels a pending default refresh when the operator stops or changes ...", failed once on the Hermetic Unit Tests job for PR #2396 (run 35134019806, 2026-09-16 18:25 UTC). The PR touches only merge-approval risk code under `src/lib/lane/`, nothing in the composer or stream hooks, and the same suite passed on the two PRs merged just before it.

The assertion was `expected [ { type: 'orchestrator-send', … } ] to have a length` of a different value: an orchestrator send was recorded after the case expected the pending default refresh to be cancelled. That points at a timing assumption in the test (the cancellation racing a scheduled refresh under a loaded runner), the same class as #2260.

## Expected

The case passes reliably in isolation and under the full hermetic run with bounded workers. Fix the ordering assumption (await the cancellation or drive fake timers deterministically); do not add retries or widen timeouts.

## Acceptance

- [ ] Root cause stated in one sentence in the fix PR.
- [ ] The case passes three consecutive full hermetic runs with the bounded-worker command from #2260.
- [ ] No reclassification unless the test genuinely owns a resource.

hurttlocker:
Closes #2397.

## Mechanic
The fresh-turn-options cancellation case switched repo and released the gated operator-defaults fetch inside one `act` scope. `act` defers the render that commits the new repo path, and with it the abort effect, until the scope ends. Whenever the fetch chain settled before the 0ms timer, the send saw the old repo path and went out. That is why it failed twice on a fast CI runner for PR #2396 and never on a loaded local machine.

## Change
Test file only. The act scope closes before the gated refresh is released, and the test awaits the actual gated fetch promise before asserting that no send was recorded. Same change applied to the interrupt half for symmetry.

## Evidence
Worker report follows in a comment.


hurttlocker:
Worker evidence (Opus 5 seat, commit 00a84dd27):

- Reproduced deliberately: the load-based commands passed, but widening the case's 0ms wait to 50ms yields the exact CI failure (`expected [ { type: 'orchestrator-send', … } ] to have a length of +0 but got 1`).
- After the fix: the file with the lane tests sharing a worker passed 5 of 5 runs (229 tests each); shuffle seeds 1 through 5 passed 6 of 6 each; `npm test` 731 files, 4611 tests passed, 4 skipped, 0 failed; `npx tsc --noEmit` clean.
- The case still guards the behavior: removing the cancel-on-repo-switch code from the hook, and separately the cancel-on-stop code, made it fail with the same assertion both times.

Residual: the exact CI trigger (file distribution across bounded workers with the #2396 test file) was not reproduced locally; the reproduction relies on widening the wait.

PR #2403, opened 2026-09-16
fix: supersede an older-HEAD workspace snapshot when capturing merge evidence

### LINKED ISSUE #2271: Existing workspace snapshot blocks merge evidence capture after reviewed HEAD changes
## Mechanic

In `src/lib/workspace/workspace-materialization-retirement.ts`, `captureWorkspaceMaterializationSnapshot` reuses an existing snapshot for the workspace and rejects merge evidence when that snapshot's `headCommit` differs from the reviewed HEAD. A legitimate newer reviewed generation therefore cannot capture its HEAD, diff, merge candidate, and recovery receipt while preserving the earlier immutable snapshot.

## Expected

When an existing workspace snapshot belongs to an older HEAD, merge evidence capture preserves that immutable history and creates a new superseding snapshot generation bound to the reviewed HEAD and current merge candidate.

## Acceptance

- [ ] Detect whether an existing snapshot matches the reviewed HEAD and merge candidate before reuse.
- [ ] Preserve an older mismatched snapshot and its transition history without mutation.
- [ ] Create one new snapshot generation containing the reviewed HEAD, diff fingerprint, merge candidate, and recovery reference.
- [ ] Make retries idempotently resolve the same current generation instead of creating duplicates.
- [ ] Add a real-path merge test that begins with a pre-existing snapshot for an older HEAD and proves the newer reviewed evidence is retained before cleanup.

hurttlocker:
Closes #2271.

## Mechanic
Merge evidence capture refused whenever the workspace already held a snapshot for an earlier HEAD, with no path to a superseding generation. Once a packet's reviewed HEAD moved after its first snapshot, the merge was blocked on "Workspace snapshot no longer identifies the reviewed merge HEAD."

## Change
- A snapshot is reused only when it certifies the reviewed HEAD and, when one was recorded, the merge candidate.
- Otherwise a materialized snapshot advances to a new generation through the existing `beginWorkspaceSnapshotGeneration` compare-and-swap, with a per-generation recovery ref. The older generation stays in the append-only receipt chain. A non-materialized snapshot still refuses.
- A postcondition re-checks certification after the write, so a lost compare-and-swap for the same evidence is reused and anything else refuses.
- A snapshot belonging to a different packet than the managed lane refuses.

## Evidence
- `tests/merge-evidence-superseding-snapshot-real-path.test.ts` (real repo registration, real git, real merge dispatch): red on unchanged source, 0 of 2 (both cases hit the old refusal). Green: 2 of 2 in the integration gate, 88s. Merge case: capture at A, commit B, merge; generation 2 has B and a recovery ref at B, generation 1 keeps A's fingerprint and ref, main has B. Idempotency case: two captures at A and two at B yield exactly two generations.
- Related suites green: replacement-owned-launch 2/2, managed-cleanup-missing-directory 17/17, retired-workspace-lane-recovery 7/7, thin-workspace-review-merge 3/3.
- `npx tsc --noEmit` clean. `npm test`: 732 files passed, 3 skipped; 4613 tests passed, 4 skipped. Manifest regenerated.

## Residual
- `tests/worktree-side-merge-real-git.test.ts` times out on the build machine with or without this change under load; main CI is green on it and that file never registers a repo, so capture returns before the changed lines. CI's integration job is the arbiter.
- The new test does not assert the final snapshot state or worktree deletion because worktree removal was refused locally by an inconclusive live-process probe. Its timeouts are 180s for the same load reason.
- A capture at the same HEAD with a different recorded merge candidate also creates a new generation, per the issue's match-both rule.


hurttlocker:
Reviewer pass: no blockers. The postcondition throws before any fast-forward, so a wrong-HEAD snapshot cannot reach the merge; retirement and cleanup key off the single current row, so multiple generations need no new handling. Residual, not a defect: the merge candidate is recomputed per attempt, so a retried merge for the same reviewed HEAD with a different candidate mints a new generation and recovery ref even though the reviewed content is unchanged. Generation churn on retries, worth a follow-up if it shows up in practice.

Issue #2430, opened 2026-09-17
syncOrchestratorControlPlaneState(state) replaces persisted state when the lock is clean, so a caller passing an older snapshot can still lose updates


hurttlocker:
Found while reviewing the #2399 fix.

## Mechanic

`syncOrchestratorControlPlaneState(state?)` in `src/lib/orchestrator/control-plane.ts` accepts an optional caller-supplied mission state. Under #2399, the three-way merge in `persistMissionUnderLock` only runs when the lock was compromised. When the lock is acquired cleanly, the caller's `state` is reconciled and written as-is. The `base` used for that write is read from disk under the lock, not what the caller read to derive `state`, so a caller that built `state` from an earlier snapshot overwrites whatever landed in between. That is the same lost-update shape #2399 closed for the compromised-lock case.

Today no caller passes `state` (checked: mobile activity and packets routes, orchestrator review-state and state routes, the operator mission merge service, and tests), so this is a latent trap on an exported function, not a live bug.

## Expected

Either the `state` parameter is removed so every caller mutates under `withLockedState`, or a caller-supplied `state` must come with the snapshot it was derived from and is always merged against the persisted state, never written blind.

## Acceptance

- [ ] No exported control-plane write path accepts a mission snapshot without the base it was derived from.
- [ ] A real-path test passes an older snapshot through the remaining public write path and asserts a packet added on disk in between survives.
- [ ] `tests/control-plane-lock-slow-holder-real-path.test.ts` and the existing lock suites stay green.

PR #2431, opened 2026-09-17
fix: wait on a live control-plane lock holder instead of writing past it

### LINKED ISSUE #2399: Control-plane cross-process lock proceeds without the lock after its wait budget, so a slow holder can still lose a state write
Found while fixing #2351.

## Mechanic

`src/lib/orchestrator/control-plane.ts` (the `acquireFsLock` loop) waits for the cross-process lock directory and, once the wait exceeds its budget, logs `cross-process lock wait exceeded budget — proceeding without lock` and continues. The comment says a lost update is recoverable by the ancestry reconciler and a wedged control plane is not.

That trade is reasonable for a deadlock, but it means two writers can run a read-modify-write on mission state at once whenever the holder is slow rather than dead: a long `withLockedState` callback in the other process, a paused debugger, or a loaded machine. The write that lands second overwrites the first. #2351 was one instance of a lost packet from an overwrite, fixed at the merge layer; this path can reproduce the same loss for any packet-level change (a status update, a stop, a receipt) that the merge does not specifically protect.

## Expected

Either the fallback does not overwrite: after the budget, the writer re-reads the current state inside its callback and merges rather than replacing, and records a lane or control-plane event that the lock was bypassed. Or the wait is bounded by the holder's liveness (the lock meta already stores `pid` and `at`), so a live holder is waited on and only a dead one is broken.

## Acceptance

- [ ] A test holds the lock from a child process past the budget while the parent writes; the child's write is not lost and the bypass is recorded.
- [ ] The existing stale-lock recovery (dead holder) still breaks the lock.
- [ ] The ancestry reconciler claim is either demonstrated for a packet-level overwrite or removed from the comment.

hurttlocker:
Closes #2399.

## Mechanic
`acquireFsLock` waited up to 8 seconds for the cross-process lock directory, then logged and proceeded without the lock. Any lock older than 10 seconds was also treated as crashed. A slow but alive holder (a long `withLockedState` callback in the other process, a loaded machine) and the new writer therefore both ran read-modify-write on `orchestrator-state.json`, and the second write replaced the first. #2351 was one instance at the merge layer; any packet-level change could hit it.

## Change
- The lock record stores pid, timestamp, and a unique token.
- The wait is bounded by the holder's liveness. A dead pid's lock is broken at once. A live holder is waited on past the 8-second budget (recorded as a slow-holder event) up to a 60-second hard ceiling, and only then broken and flagged.
- A process releases the lock only while it still owns it, so a broken lock is never removed by its old holder.
- When a lock was compromised, `persistMissionUnderLock` re-reads the persisted state and does a three-way merge (packets the other writer added or changed survive; fields this writer did not change take the persisted value) instead of replacing. Break, slow wait, and merge are appended to `control-plane-lock-events.jsonl` in the data dir.
- The comment claiming the ancestry reconciler recovers a lost update is removed; it reconciles merged lanes and never restores dropped packet rows.
- `O8_CONTROL_PLANE_LOCK_WAIT_BUDGET_MS` and `O8_CONTROL_PLANE_LOCK_HARD_CEILING_MS` exist so tests can shorten the timings.

## Evidence
- New real-path test `tests/control-plane-lock-slow-holder-real-path.test.ts` spawns a real child process that holds the lock through `withLockedState` while the parent writes through the same function: (1) child holds 1.5s against a 300ms budget, both packets persist, one slow-holder event names the child pid; (2) child holds past an 800ms ceiling and writes after the parent, both packets survive, the break names the child pid, the merge event lists the parent's packet as preserved; (3) a lock left by a dead pid with a fresh timestamp is broken in under 2 seconds.
- Fail-before against the previous `control-plane.ts`: `expected [ 'pkt-child' ] to deeply equal [ 'pkt-child', 'pkt-parent' ]` and `expected 8018 to be less than 2000`, 0/3 passing.
- With the change: new test plus `mission-stop-lock-real-path` and `retry-salvage-lock-reentry-real-path`: 3/3 files, 9/9 tests. `npm test`: 739 files, 4678 tests passed, including the existing dead-holder unit test. `npx tsc --noEmit` clean. New file classified child-process, process-signal, real-path.
- Independent read-only review: no path breaks a live holder before the ceiling, no path leaves a dead holder unbroken, the compromised check has no false negative in the break-and-retake window, and the test fails on the old code.

## Residual
- `withControlPlaneLock` callers write for themselves, so a broken lock there records a `bypass_unmerged` event and nothing more.
- Same-packet edits by both writers keep the bypassed writer's version; two packets added at once can share a `referenceLabel`.
- A hung live holder now stalls writers up to 60 seconds where the old path gave up at 8. The event log is append-only without rotation.
- `syncOrchestratorControlPlaneState(state)` writes a caller-supplied snapshot blind when the lock is clean; no caller passes one today. Tracked in the follow-up issue filed alongside this PR.


hurttlocker:
Follow-up for the residual on the caller-supplied snapshot path: #2430.

PR #2477, opened 2026-09-18
fix: reap non-git packet directories once instead of retrying every tick (#2474)

### LINKED ISSUE #2474: Worktree reaper retries non-git packet directories every 5 minutes and floods the log
## Mechanic
`src/lib/lane/worktree-reaper.ts` runs a terminal sweep every 5 minutes. For a packet worktree directory whose `.git` file is gone (the directory still exists with agent scratch files in it), the sweep calls the cleanup path in `src/lib/lane/worktree-cleanup.ts`, which runs `git status --porcelain` to preserve the head before removal. Git answers `fatal: not a git repository: (null)`, the capture and preserve steps both log a warning, the removal is counted as failed, and the directory stays. Nothing marks the entry as unrecoverable, so the same three directories fail again on every tick.

## Observed
On the installed app, one repo's `.cortex-worktrees` holds three such directories dated a week ago. Every 5 minutes the app log gains ~10 warning lines for them and one summary line:

```
[worktree-reaper] terminal sweep repos=17 scanned=5 removed=0 skippedActive=2 failed=3
```

30 occurrences in the current (25 KB) log file alone. The log rotates before any real signal survives, and the directories never go away.

## Expected
A terminal-sweep target that is not a git repository has no head to preserve. The sweep should treat that answer as "nothing to capture", remove the directory (it belongs to a terminal lane by the sweep's own selection), and log once. If removal is refused for another reason, the entry is recorded as unrecoverable so later ticks skip it instead of retrying and re-logging.

## Acceptance
- Real-path test through the reaper's sweep with a fixture repo whose `.cortex-worktrees` contains a plain directory (no `.git`) for a terminal-state lane: after one tick the directory is gone, `removed` counts it, and the log carries one line for it. A second tick logs nothing for it.
- A directory that fails removal for a non-git reason is skipped on the next tick without new warnings.

hurttlocker:
Closes #2474.

## Mechanic
The 5-minute terminal sweep reaches packet directories under a repo's `.cortex-worktrees` whose git link is gone. The cleanup path's preserve-head step ran git there, got "not a git repository", logged a block of warnings, counted the removal as failed, and left the directory in place. Nothing marked it unrecoverable, so the same three directories re-failed on every tick and the log filled with the same block (30 blocks in one 25 KB log file on the installed app).

## Change
- `src/lib/lane/worktree-cleanup.ts`: in the terminal/force path, check first whether the target is a git repository (no `.git` entry, or `git rev-parse --git-dir` answers not a repository). If it is not, there is no head to preserve: remove the directory behind the existing live-process guard, settle its storage reservation, and log one line. The check runs before any git call, so git never walks up and answers for the parent checkout.
- `src/lib/lane/terminal-worktree-sweep.ts`: a directory whose removal failed is remembered for this process (path plus lane id and status) and skipped on later ticks, counted as `skippedUnrecoverable`. A lane status change or a restart retries it.
- `src/lib/lane/worktree-reaper.ts`: the summary line carries the new count.

## Evidence
- `tests/worktree-reaper-non-git-dirs-real-path.test.ts` drives the real maintenance tick against a temp repo whose `.cortex-worktrees` holds a non-git directory for a terminal lane: after one tick the directory is gone and counted as removed with one log line; the second tick logs nothing for it. A removal that fails for a non-git reason is skipped on the next tick without new warnings.
- Fail-then-pass: the new test against origin/main reports 2 failed, 1 passed; with the fix, 3 passed. Integration gate 2/2 files, 5/5 tests. `npm test`: 747 files passed, 3 skipped. `tsc` and eslint clean.
- The existing preservation test's second-sweep expectation moved from a repeated failure to `failed: 0, skippedUnrecoverable: 1`; its first-sweep head-banking assertion is unchanged.

## Residual
- The live `(null)` error came from a `.git` file pointing at a deleted gitdir, not from a missing `.git`. A directory with no `.git` entry at all previously passed the preserve step because git walked up and answered for the parent checkout; the explicit check now stops that before any git call.
- A directory refused by the live-process guard is not retried until its lane's status changes or the app restarts. The unrecoverable set is in memory by design.


hurttlocker:
Independent review found one regression on 7ab18eac0: the skip-on-later-ticks set was fed by any removal failure, so a live-process-guard refusal (transient by design) would be skipped for the lane's remaining lifetime at that status. A follow-up commit scopes the set to the non-git classification and adds a third-tick recovery case to the test. Holding merge until it lands and is re-reviewed.

hurttlocker:
Follow-up commit d879f0cef: the skip set is now fed only when the failed directory still exists and is not a git repository; every other failure, including a live-process-guard refusal, retries each tick as before. The preservation test's second-sweep expectation is back to a retried failure, and the non-git test gained a live-guard case on a valid clone (retries on ticks 1 and 2, removed on tick 3 once the guard clears). Re-review in progress.

Issue #2445, opened 2026-09-17
Tracking: referee-ranked mobile surfaces (faster operator decisions on the phone)


hurttlocker:
Sibling of #2433 (the referee seam). Everything here runs on the desktop and ships results to the phone; the phone never calls the referee and no key leaves the desktop.

## Outcome

The phone gets faster where it is actually slow: the operator deciding what to tap, and the desktop deciding what to send. Approvals with trustworthy facts become one tap. The inbox is ordered by need, not age. Pushes fire for things that need a person. Phone Symon starts with the right tools, can watch fuzzy conditions, and briefs the important change first.

## Rules (inherit from #2433)

Advisory only, per-question answers only, thresholds only from the local calibration replay (#2438), setting default off with today's behaviour as the fallback, a receipt for every call, no worker-written text in the state.

## Children

- [x] 2439 One-tap approve chips on phone inbox cards (#2478, 335e286bb)
- [x] 2440 Referee-ordered mobile inbox (#2471, 2c6de76ec)
- [ ] 2441 Push notification gate
- [ ] 2442 Phone Symon first-utterance tool-pack routing
- [ ] 2443 Symon watches with fuzzy conditions
- [ ] 2444 Catch-up ranking

## Order

1 to 3 after #2434 and #2435 land. 4 to 6 belong with the Symon lane and can start once #2434 lands.

hurttlocker:
#2440 merged in #2471 (2c6de76ec). Desktop orders the phone inbox by a per-item urgency score when the referee setting is on; payload byte-identical when off; the route never waits on the provider. Follow-up #2475 (cache fingerprint should include merge-card referee facts). Next child: #2439 approve chips, which reads the stored merge-card facts from #2435 plus the docsOnly threshold.

hurttlocker:
#2439 merged in #2478 (335e286bb). One calibrated chip today, Docs only, shown when o8's path rule and the stored referee answer (at or above 0.9) agree; approving from a card with a chip goes through the same panel approvals call as every other card, adding a recorded fact on the event. Follow-ups filed from review: #2475 (urgency cache fingerprint), #2479 (mobile action route drops merge strategy and stamps a different actor; pre-existing), #2480 (equality test on a merge that completes). Not yet checked on an iPhone; that happens before the next release. Both #2440 and #2439 are on main, unreleased.

hurttlocker:
[...]

Issue #2440, opened 2026-09-17
Referee-ordered mobile inbox: urgency score computed on the desktop


hurttlocker:
Child of #2445. Depends on #2434.

## Mechanic
The mobile inbox orders pending items by age. Under load the operator scrolls past routine cards to find the one that is blocking a lane.

## Expected
The desktop asks the referee one score question per pending item ("how urgently does this need the operator?") over the item's facts (kind, lane state, age, risk, whether a lane is blocked on it) and ships the phone an order. Ties and abstains keep age order. The score and its receipt are visible in the item detail. Setting off: age order as today.

## Acceptance
- [ ] Real-path test through the mobile inbox route with the local endpoint fixture: items come back in score order with the score attached; with the setting off the order is unchanged.
- [ ] A referee failure or abstain on one item does not reorder the others or delay the response beyond the retry budget.
- [ ] The score question text lives in the locked questions file.

Issue #2435, opened 2026-09-17
Referee row on approval cards, advisory beside the rule risk


hurttlocker:
Child of #2433. Depends on the judgment client child.

## Mechanic

An approval card shows a risk level derived from changed paths (#2381) and whatever the gate recorded. It does not show any read of the diff content itself. The operator decides from the title, the summary, and the diff.

## Expected

When `judgment.provider` is on, creating an approval that carries a diff also asks the referee six locked questions and stores the receipt on the approval: docs-only, adds tests, tests reach a real entry point, touches middleware or auth, contains placeholder or mock data, substantial changes the title and summary do not account for, plus a 5-level risk score. The card renders a "Referee" row with each answer as a probability and the model string, next to the existing rule risk. Nothing about the decision path changes: no auto-approve, no threshold, no reordering. When the setting is off or the call returns null, the row is absent and the card is exactly as today.

## Acceptance

- [ ] Real-path test through the real approval creation path used by merge cards (the same path #2381's test drives) with the local endpoint fixture: the approval row carries the receipt and the API response for the card includes the referee answers.
- [ ] With the setting off, the same path produces a byte-identical approval to today's.
- [ ] Referee failure (null) never blocks or delays approval creation beyond the bounded retry budget; test asserts the approval exists before the fixture responds.
- [ ] Card renders the row on desktop with inline styles and theme tokens; no new CSS classes; screenshot or view read attached to the PR.
- [ ] The rule risk stays the value shown as "Risk"; the referee's score is labeled as advisory.

hurttlocker:
Claimed. Building on a branch cut from main after #2434 landed; PR to follow with real-path tests.

Issue #2434, opened 2026-09-17
Judgment client with receipts, setting, and diff sanitizer


hurttlocker:
Child of #2433.

## Mechanic

Nothing in o8 can ask a typed yes/no, pick-one, or score question of a judgment model and keep a receipt of the answer. Every place that wants one (approval cards, Brain routing, rerun warnings) would otherwise build its own call, its own key handling, and its own logging.

## Expected

One module, `src/lib/judgment/`, that the other children call:

- `askJudgment({ state, questions, context })` posts to the configured provider (first provider: the TypeSafe systemone endpoint, model `jev-latest`, direct fetch, no SDK) and returns typed answers: noul (0..1), choice (choice, probabilities, confidence), score (score, legend, probabilities, confidence). Returns null when the setting is off, the key is missing, or the call fails after bounded retries on 429/5xx. Never throws.
- Setting `judgment.provider`: `off` (default) or `typesafe`. Read through the operator settings store like other operator defaults; surfaced in Settings with copy that says diff content is sent to the provider when on.
- Key from `~/.o8/judgment-api-key` (0600) or the `O8_JUDGMENT_API_KEY` env var; never logged.
- Receipt for every call, including failures: question ids and exact text, answers, provider model string, input and output tokens, latency ms, truncation flag, context (packetId, laneId, approvalId, surface). Stored as a `judgment` lane event when a lane is in context, otherwise in a small `judgment_receipts` table.
- `buildDiffState(files, diffText, budgetTokens)`: file list with additions and deletions computed by o8, path-derived `docsOnly`, then the diff with commit-trailer lines (`Reviewed-by`, `Signed-off-by`, `Approved-by` and the like inside added lines) and lines addressed to a reviewer removed, cut to the budget with whole small files kept and head plus tail of large files, truncation marked. Budget default 24K tokens at a measured 3.2 chars per token for code.
- Question texts live in one file with a comment per question; wording is treated as an API and changed only with a calibration rerun.

## Acceptance

- [ ] Real-path test: with the setting off, `askJudgment` returns null without opening a socket (assert via a fetch spy that is never called).
- [ ] Real-path test against a local HTTP fixture that mimics the endpoint: answers are typed, a receipt row exists with the exact question text and usage, a 429 then 200 sequence succeeds, a persistent 5xx returns null and records the failure receipt.
[...]

PR #2478, opened 2026-09-18
feat: docs-only approve chip on phone inbox cards from stored referee facts (#2439)

### LINKED ISSUE #2439: One-tap approve chips on phone inbox cards from referee facts
Child of #2445. Depends on #2434 and #2435.

## Mechanic
The slowest step on the phone is the operator reading a diff on a small screen before approving. The card carries a title, a summary and a risk word; nothing says "this is docs only" or "this adds tests" in a way the operator can trust without opening the diff.

## Expected
When the desktop has referee facts for an approval (docs-only, adds tests, risk score) above the thresholds set by the local calibration replay, the mobile inbox card shows a chip per fact ("Docs only", "Adds tests") and the approve action is reachable from the card without opening the diff. Below threshold or with no facts, the card is unchanged. The chip carries the probability on long-press and the receipt is linked from the card.

## Acceptance
- [ ] Real-path test through the mobile inbox route: an approval with a stored receipt above threshold serialises the chips; one below threshold or with no receipt does not.
- [ ] Approving from the card records the same approval event as approving from the diff view, plus the fact that chips were shown.
- [ ] Setting off: byte-identical inbox payload.

hurttlocker:
Closes #2439. Child of tracker #2445. Depends on #2434, #2435, #2440 (all merged).

## Mechanic
The slowest step on the phone is reading a diff on a small screen before approving. The desktop already stores the merge-card referee facts on each approval (#2435). This surfaces one calibrated fact on the phone card as a chip, and lets the operator approve from that card, recording that the chip was shown.

## Change
- `src/lib/judgment/questions.ts`: `DOCS_ONLY_CHIP_THRESHOLD = 0.9`. Calibration recorded in the comment: on the 118-diff lab run every path-docs-only diff scored 0.97 to 0.98 and every other diff 0.01 or lower; under a lying-title attack (titles are now excluded from state) docs diffs fell to 0.79 to 0.85. Recalibrate with the replay before changing.
- `src/lib/mobile/inbox-referee-chips.ts` (new): a "Docs only" chip is attached only when o8's own path-derived docsOnly is true (the rule in `judgment/diff-state.ts` over the approval's diff) AND the referee's docsOnly answer is at or above the threshold and not an abstain. Setting off, no stored facts, or below threshold: the item keeps its exact bytes. No other chip is shown; addsTests has no replay yet and risk has no gate signal on this install.
- `src/lib/mobile/inbox.ts`: chips are attached in the same last step as urgency ordering (#2440), outside the snapshot cache, because the referee stores its facts after the card is created.
- Approve from the card: the existing mobile approve action accepts `via: 'chip'`; unknown values are ignored. The approval event gains `approvedFromCard: { via: 'chip', chipsShown }`, with the chip kinds recomputed on the desktop, never taken from the phone (`approvals/resolution.ts`, `approvals/types.ts`, `api/mobile/action/route.ts`).
- Phone UI: `components/mobile/InboxRefereeChips.tsx` renders the chip on the approval card (44px target, raw SVG, card palette); long-press reveals the probability and receipt id as selectable text. A card with a chip approves through the mobile action with `via: 'chip'`; cards without chips are unchanged.

## Evidence
- `tests/mobile-inbox-referee-chips-real-path.test.ts` through the real `/api/mobile/inbox` and `/api/mobile/action` routes against persisted approvals: 1/1 PASS, 4/4. Chip present above threshold with docs paths; absent below threshold, with a non-docs path, and with no facts; setting off byte-identical; approve with `via: 'chip'` persists the event fact, plain approve is unchanged.
- Fail-then-pass: with the chip step removed, 2 of 4 fail.
- #2440 and #2435 suites: 2/2 files, 8/8 tests. `npm test`: 747 files passed, 3 skipped; 4726 tests passed. `tsc` and eslint clean.

## Residual
- Not checked on a device yet. The phone UI should be looked at on an iPhone before release.
- Chip cards approve through `/api/mobile/action`; non-chip cards keep the pre-existing `/api/panel/approvals` path. The reviewer compared the two persisted outcomes before merge.
- No receipt viewer exists on the phone, so the receipt id is shown as text rather than a link.
- Three touched files were already over the 800-line ceiling before this change and grew by 2 to 5 lines each.


hurttlocker:
Independent review found a blocker on e4ae47fd6: chip cards approved through the mobile action route while every other card uses the panel approvals route, and the two diverge for lane-merge continuations (stored strategy not forwarded; different audit actor). The test also seeded approvals without a continuation, which is not the shape that earns chips. A follow-up commit routes chip cards through the same path as every other card and seeds the real lane-continuation shape in the test. The underlying route divergence is filed separately. Holding merge until the follow-up lands and is re-reviewed.

hurttlocker:
Follow-up commit 921df302e: chip cards now approve through the same panel approvals call as every other card, adding only `via: 'chip'`; the panel route validates it and records the event fact. The separate mobile-action path and its `via` field are removed, so that route matches main again (its own divergence is #2479). The test now seeds real lane-merge approvals with a non-default strategy and stored referee facts, and asserts plain and chip approvals dispatch the same lane command, strategy included, with matching status, actor, and event fields apart from the recorded fact. Re-review in progress.

Issue #2481, opened 2026-09-18
Tracking: typed judgment referee across o8 (program home)


hurttlocker:
Successor to #2433 (closed 2026-09-17 with all five children merged). This is the standing home for the typed judgment referee across o8. Label: `area:judgment`. Milestone: Judgment referee.

## What it is
A non-generative judgment model answers locked, typed questions over state that o8 computed itself (git-derived file lists, diff text, lane state, path facts). o8 never asks it what to do; it asks per-question facts, records a receipt for every call, and reads a threshold only where the local calibration replay earned one. Everything is behind `judgment.provider`, default off. Bring-your-own key works on every plan today.

## Rules that hold everywhere (from the 2026-09-17 lab and replay)
- State carries o8-computed facts and diff text only. No worker-written title, summary, or trailer reaches the referee.
- Per-question answers only. `recommendedAction` is record-only, never read.
- Thresholds only where a replay on this install's own history earned them. Today: `docsOnly` (0.9 for the phone chip). Risk score has no gate-failure signal on this install (AUC 0.54); it stays advisory.
- Every call writes a receipt the operator can read. Abstain below 0.4 confidence; the human decision path is unchanged.
- Replay before any gate. Nothing auto-decides.

## Shipped (v0.1.758)
- [x] #2434 judgment client, receipts, setting, diff sanitizer
- [x] #2435 referee row on approval cards, advisory beside the rule risk
- [x] #2436 Brain classifier: typed Class A / Class B as the first tier
- [x] #2437 gate-failure early warning recorded before the automatic rerun
- [x] #2438 calibration replay over local approval history
- [x] #2459 receipts migration reconciled on old-shape tables

## Shipped (v0.1.759, 2026-09-18)
Every surface below is advisory and writes a receipt. None of them gate anything.
- [x] #2440 referee-ordered phone inbox (urgency score computed on the desktop)
- [x] #2439 docs-only approve chip on phone cards, approve from the card with a recorded fact
- [x] #2441 push notification gate, detached from the send
- [x] #2443 Symon watches with fuzzy conditions
- [x] #2444 catch-up ranking, and #2511 the receipt identity that lets it be scored
- [x] #2446 advisory rule citations on the merge preview
- [x] #2447 claim-versus-evidence check on the worker's final report
- [x] #2448 loop and stall detection on the supervisor tick
- [x] #2465 judgment-scored compaction entries, record-only
- [x] #2467 event triage before an orchestrator wake
[...]

Issue #2448, opened 2026-09-17
Loop and stall detection from tool-call patterns, not just missing heartbeats


hurttlocker:
Child of #2433, later wave. Depends on #2434.

## Mechanic
The supervisor notices a worker that stops sending heartbeats. It does not notice a worker that keeps working in a circle: the same failing command, the same edit undone and redone, the same file read ten times. Those packets burn tokens until a human looks.

## Expected
When the setting is on, the supervisor tick asks one yes/no question per active packet over the last N tool calls ("is this agent repeating the same actions without progress?"). Two consecutive positive ticks above the calibrated threshold raise a `possible_loop` lane event and an inbox item with the repeated pattern quoted. Advisory: nothing is stopped automatically. Legitimate retry-with-fix loops are the false-positive class the replay must measure before any threshold is set.

## Acceptance
- [ ] Real-path test through the supervisor tick with the local endpoint fixture: a transcript window with the same failing command five times raises the event on the second positive tick, not the first; a window of distinct commands does not.
- [ ] The inbox item shows the repeated calls and the receipt id.
- [ ] Setting off: tick unchanged.

hurttlocker:
Landed in #2506 (b85017bd6). Unreleased: it reaches the installed app at the next ship. Recording only, advisory, and it stays that way until the promotion gate in #2481 is satisfied for this surface.

PR #2506, opened 2026-09-18
feat: record loop checks on the supervisor tick and raise an advisory possible_loop (#2448)

hurttlocker:
Refs #2448. Program #2481, wave 1. RECORD-ONLY and ADVISORY: nothing is stopped, steered, or gated.

## Mechanic
- New `src/lib/supervisor/loop-detector.ts`. On each supervisor tick, when the judgment provider setting is on, `startLoopChecks` runs a detached check per running agent bound to an active packet lane: at most once a minute per lane and only after a new tool call landed. The tick never awaits it.
- State is o8-computed facts from the last 20 tool calls: tool name, sha256 of the normalized call text, the first 200 chars of the call's output, an error flag, and counts (distinct tools, distinct argument hashes, repeated runs). No assistant text, packet title, or worker report.
- Every answer is a `loop_check` lane event. Two consecutive answers at or above the PROVISIONAL band (0.6) raise one `possible_loop` lane event and one inbox item per run, quoting the repeated call and the receipt id. An answer at or below 0.4 clears the streak.
- Inbox: new `possible_loop` kind through the real supervisor inbox store; copy states it is advisory and nothing was stopped.
- Replay: `node scripts/judgment-replay.mjs --label loop` reads `loop_check` events read-only and sends nothing.

## Tests (real path, `tests/loop-detector-real-path.test.ts`)
Persisted lane and watched-agent rows, the real supervisor tick, provider setting through the real operator-defaults store, the call over HTTP to the local endpoint fixture, events read back from SQLite.
- tick 2 raises exactly one `possible_loop` plus an inbox item; tick 3 still one
- distinct commands: zero `possible_loop`
- request body carries neither the assistant text nor the packet title
- setting off: zero fixture requests, tick events identical to baseline
- replay prints n, AUC, Brier
- fail-then-pass: with the hook removed, line 199 fails with `expected [] to have a length of 1 but got +0`

New file 6/6; `npm test` 752 files / 4753 tests passed; `tsc` 0; lint clean on touched files.

## Residual
- The supervisor's transcript callback drops tool arguments, so the hash is over the tool entry's text.
- Dispatched Claude Code sessions do emit `tool` entries (`src/lib/runtimes/shared/owned-transcript.ts`), so both runtimes are checked. An earlier draft of this description said otherwise; that was wrong.
- `agent-supervisor.ts` was already past the 800-line ceiling (1161); this adds 8 lines for the hook and a test entry point. Decomposition is a separate change.


hurttlocker:
Independent review correction: the description originally listed "Claude Code workers produce no tool entries and are not checked" as a residual. That was wrong. Dispatched (owned) Claude Code sessions emit `role: 'tool'` entries through `src/lib/runtimes/shared/owned-transcript.ts`, fed by `src/lib/claude-code/owned.ts`, the same as Codex sessions in `src/lib/runtimes/codex.ts`. Packet lanes only exist for owned launches, so both runtimes reach the check. The body has been corrected.

PR #2548, opened 2026-09-20
Fix MCP mission effort pin

### LINKED ISSUE #2547: Operator mission creation drops explicit worker effort
## Mechanic

The operator mission-creation surface accepts an explicit worker model but does not preserve an explicit reasoning effort. In a staged mission, `requestedEffort` and `selectedEffort` both remained null after requesting medium. A subsequent launch would therefore use the configured worker default instead of the requested pin.

## Expected

An explicit supported worker effort is exposed by the operator tool contract and remains intact through mission creation, persisted routing, and dispatch. Invalid values must not silently fall back. Omitted effort retains existing behavior.

## Acceptance

- [ ] Expose and validate explicit effort on the operator mission-creation entry point.
- [ ] Persist the requested effort and prove dispatch selects the same supported value.
- [ ] Cover invalid and omitted effort without altering global settings.
- [ ] Add a real-entry-point regression with persisted state and a fake launch boundary; record failure before the fix and success afterward.

Related compatibility audit: #2520.

hurttlocker:
Fixes #2547

## Mechanism
The operator MCP create_mission tool now exposes and validates the effort pin, then forwards it through both GitHub-backed and inline mission transport paths to the existing create-mission route.

## Proof
- Real path: MCP tool -> persisted packet routing -> dispatch -> fake Codex argv preserves `medium` instead of the `xhigh` default.
- Omitted effort remains runtime-default behavior; malformed or conflicting values fail before mission creation.
- `npx tsc --noEmit`, touched-file ESLint, rule-check, focused tests, and `npm test` passed (4,780 tests, 4 skipped).

## Boundary
No live provider/model calls, deployment, merge, or settings changes.


hurttlocker:
Independent review of `83070954e54aade264f2f5f704b8e2291fa40620`: no blocking findings in the three-file diff.

Regression check against the previous production implementation (`7ec08a7e4e13f2c6c191bd4088b0ef4fe34ab880`), keeping the completed tests:

- The explicit-medium MCP case fails because persisted `requestedEffort` and `selectedEffort` are both null rather than medium.
- The malformed-effort case fails because the old handler accepts the invalid value instead of rejecting it.
- The fixed implementation passed the full 21-test effort suite and the 4,780-test unit gate.

The checks use isolated persisted state and a fake executable. Installed-app verification remains separate from this source-level proof.

PR #2555, opened 2026-09-20
Fix owned worker transcript grouping

### LINKED ISSUE #2549: Worker transcript renders streamed fragments as separate messages
## Mechanic

The installed worker transcript panel renders fragments of one assistant response as separate chat messages. Each fragment receives its own divider, action controls, and timestamp. Markdown spans are split across rows, leaving unmatched backticks and broken inline code. The adjacent orchestrator transcript groups complete responses normally.

Observed in the live desktop app as well as an operator-provided screenshot. The ingestion and message-boundary cause is still under investigation.

## Expected

Worker transcripts preserve complete message boundaries and render Markdown consistently with the orchestrator transcript. Streaming updates extend the correct message; separate turns, tools, and reasoning retain their proper boundaries.

## Acceptance

- [ ] Reproduce the observed fragmented worker history through the real transcript entry point.
- [ ] Preserve complete assistant text and Markdown without merging unrelated messages or duplicating snapshots and deltas.
- [ ] Apply the same grouping rules to history load, live updates, and reconnect.
- [ ] Verify the rendered worker panel against the orchestrator panel using representative retained events.
- [ ] Capture before/after visual proof in a dev preview for operator review before release.

hurttlocker:
## Mechanic
Owned Claude Code stream deltas now update stable content-block entries instead of rendering one row per fragment. Thinking and tool events remain separate, and a terminal result does not replay streamed answer text.

## What changed
- Preserve block identity and whitespace through the parser and owned transcript projection.
- Let incremental readers replace the current Claude entry by stable ID.
- Add a persisted JSONL route regression for split Markdown, thinking, tools, terminal replay, and incremental updates.

## Verification
- `npx vitest run src/app/api/mobile/history/route-runtime-parity.test.ts src/lib/claude-code/stream-json-parser.test.ts src/components/desktop/workspace-terminal/workspace-transcript-composition.test.ts`
- `npx tsc --noEmit`
- Touched-file ESLint
- `npm run rule-check -- --base=origin/main`
- `npm test`

Fixes #2549


hurttlocker:
Review held pending the queued correction:

- A content-block index is only unique within a provider message. Reusing index 0 after a tool round currently merges distinct messages and can move later content ahead of intervening tools. Preserve the message identity as well as block identity.
- Streamed thinking followed by the complete assistant snapshot currently appears twice. The real dev transcript reproduces this, so replay deduplication needs a regression case.
- Verify tool results stay associated with their tool call rather than appearing as standalone prose.

The existing passing tests do not cover these sequences. This PR is not accepted for release yet.

hurttlocker:
The transcript now groups streamed text by provider message and content block, preserves Markdown whitespace, keeps reasoning in collapsed cards, and pairs tool results with their original tool rows. Incremental reads replace changed rows by stable ID, including out-of-order tool completion before any answer text. Terminal errors remain visible without duplicating an already rendered error.

Validation: persisted history and parser regressions (21 passed), genuine red-to-green reproductions for tool-only cursor updates and repeated terminal errors, touched-file lint and rule-check. Independent final review found no remaining concrete issue. A retained session was checked through the live dev renderer: complete formatted answers, collapsed reasoning, structured tools, and one terminal error. Full typecheck and unit-suite results are recorded separately when complete.

This supersedes the earlier incomplete acceptance evidence. No release has been published.

Final local completion: TypeScript passed, full hermetic suite passed (4,794 tests; 4 skipped), lint and changed-line rule checks passed. Review candidate: `31be7b2dc`.

PR #2561, opened 2026-09-20
fix: anchor composer menus across UI zoom

### LINKED ISSUE #2559: fix: keep composer popovers anchored during UI zoom
Composer model and effort menus drift away from their trigger when the application zoom changes. Keep the shared composer popover in the same coordinate space as the scaled document, respond to zoom while open, and constrain it to the visible viewport. Acceptance: open model and effort controls, use zoom-in and zoom-out shortcuts, verify trigger alignment and viewport containment without closing the menu. Preserve existing scroll, resize, keyboard, and dismissal behavior.

hurttlocker:
Open composer menus drifted away from their buttons when UI zoom changed because screen rectangles and layout dimensions used different scales. The shared popover now places the panel in document coordinates, observes root zoom changes while open, and keeps placement within the viewport. Escape handled by an inner control no longer also dismisses the whole menu.

Validation: four component regressions fail on the original implementation and pass with the fix. Live dev checks used the actual zoom shortcuts at 90%, 110%, and 125%; the trigger and panel right edges stayed aligned. Independent review, typecheck, changed-file lint, and rule checks passed. Full-suite result is recorded below.

Fixes #2559


hurttlocker:
Full local completion gate passed: 4,798 tests passed, 4 skipped. TypeScript, changed-file lint, rule checks, focused regressions, and independent review are green.

hurttlocker:
CI caught a type assertion in the test fixture after its lint adjustment. The fixture now passes complete typed props without the assertion. Fresh local TypeScript, lint, and all four focused tests pass on ac4b9a556; CI is rerunning for that exact head.

PR #2562, opened 2026-09-20
fix: review exact lane changes before merging

### LINKED ISSUE #2560: fix: preview approved changes before status-bar merge
The status-bar merge action shows only a count and destination, so the operator cannot tell which changes will be merged. The first click should open a read-only preview identifying the source work, target branch, approval state, and available file/diff details; the actual merge action belongs inside that preview. Hover or keyboard focus should provide a short summary. Reuse existing review data and merge gates, and preserve failure/loading states. Acceptance: inspecting the action performs no merge, details identify the intended work, and only the explicit preview action invokes the existing merge flow.

hurttlocker:
The status bar previously offered an immediate merge without identifying the work and implied a bulk merge while selecting only the first approved lane. It now opens the exact lane for review, with a chooser when several lanes are approved. Hover and accessible labels identify the work and source branch; the review surface shows the diff and retains the explicit governed merge action.

The chooser supports keyboard dismissal and scrolling for long queues. The lane diff header displays the source and destination from the existing authoritative diff response, without treating an immutable comparison commit as a destination branch.

Rapid lane switching also discards stale diff responses and hides the prior lane data immediately, so the visible changes cannot belong to a different packet than the review action. An overlapping-response regression failed before this fix and passes afterward.

Validation: independent final review cleared the change. Browser verification used the real status control with sample lanes and no connected merge action; selection, keyboard focus, Escape, and viewport placement passed. The existing governed merge remains covered by its review-header test. Completion gate results follow below.

Fixes #2560


hurttlocker:
Local completion gate passed: 4,805 tests passed, 4 skipped, plus TypeScript, touched-file lint, rule checks, and independent review. The stale lane-response regression failed before repair and now passes. Real browser verification also caught first-focus timing while the portal was hidden; focus is now deferred until placement, and keyboard selection was rechecked. No live merge was performed during UI verification.

